PR #2, opened 2016-05-06
Add UEFI RFC from lzap

GregSutcliffe:
Submitting this on @lzap's behalf since I was a bit slow getting the repo transferred. @lzap @ohadlevy et al. - over to you ;)


text/0000-PXE-Booting-UEFI.md
```diff
@@ -0,0 +1,228 @@
...
+
+Currently, Foreman supports booting of x86 BIOS systems via PXELinux and
+several other platforms via Grub1, ZTP and POAP. These are represented as PXE
+"variants": syslinux, pxegrub, ztp and poap which are hardcoded in operating
+system model classes. However, EFI is not currently supported, and should be.
+
+# Detailed design
+[design]: #detailed-design
+
+Currently, Foreman supports booting of x86 BIOS systems via PXELinux and
```
ohadlevy: similar paragraph to the motivation part? (dry?)

text/0000-PXE-Booting-UEFI.md
```diff
@@ -0,0 +1,228 @@
...
+QEMU/KVM). The only version which was found working was 6.03-pre20.
+
+Additionally, there is no PXELinux version with EFI support in RHEL6 and RHEL7
+and both systems ship with Grub (versions 1 and 2 respectively). Grub is more
+mature when it comes to EFI support and since it can boot via PXE and Foreman
+already supports Grub1, it is more feasible to get EFI working via officially
+supported Grub. To have complete support matrix, PXELinux will be part of this
+implementation, but hardware support (and thus testing) will be limited.
+
+The first idea was to simply deploy both BIOS and UEFI configurations during
```
ohadlevy: I find this section confusion, as it looks like this is part of the solution while it actually not. imho this could be an appendix / optional or something similar but later down the document.

ohadlevy: and of course, the main reason for being against this part is that it mostly works on ISC DHCP vs others (MS DHCP etc)

text/0000-PXE-Booting-UEFI.md
```diff
@@ -0,0 +1,228 @@
...
+PXEGRUB |	PXEGrub | BIOS | Solaris-* | boot/grub/menu.lst.01aabbccddeeff | boot/grub/menu.lst
+POAP | POAP |	- |	poap.cfg/aabbccddeeff | poap.cfg/AABBCCDDEEFF | -
+ZTP | ZTP | - | poap.cfg/aabbccddeeff | poap.cfg/AABBCCDDEEFF | -
+
+PXELinux covers all x86 systems, PXEGrub is used for Solaris (firmware filename
+is Solaris-*-pxegrub) and POAP/ZTP have built-in loaders and only downloads
+configuration files. This feature aims for this support matrix:
+
+Variant |	Template kind | Arch | DHCP filename(s) | MAC config | Default config | Comment
+--- | --- | --- | --- | --- | --- | ---
```
ohadlevy: last column seems to be cut off?

text/0000-PXE-Booting-UEFI.md
```diff
@@ -0,0 +1,228 @@
...
+(version 1) must deploy two configuration files instead of one now to support
+both BIOS and UEFI naming conventions (grub/menu.lst vs grub/efidefault and
+grub/menu.lst.01aabbccddeeff vs grub/01-AA-BB-CC-DD-EE-FF). Also, Pxegrub
+variant will use new base directory "grub/" instead of "boot/grub/" for
+consistency with "grub2" and "pxelinux.cfg/".
+
+### [Foreman core #12634](http://projects.theforeman.org/issues/12634) - New host/hostgroup flags pxe_template and loader
+
+Host and hostgroups must be enhanced with two new flags: PXE template specifies
+PXE template kind and variant pair to use. Until today, this was hardcoded to
```
ohadlevy: can't we extend the model object instead? would set the default boot flag on the hardware model.

ohadlevy: my main point is that it doesn't feel right on the hostgroup object, maybe we could introduce another object that combines the os, template boots strategy etc.?

ohadlevy: another option, is that we deploy all kind of configuration files in one action (e.g. we could setup pxelinux and grub at the same time regardless of how the user uses it?)

text/0000-PXE-Booting-UEFI.md
```diff
@@ -0,0 +1,228 @@
...
+* PXEGrub BIOS default local boot
+* PXEGrub UEFI default local boot
+* etc. (the same for global default templates)
+
+But only kind/variant is not enough, users need to select BIOS or UEFI booting
+method. This is where Loader flag comes in - selection of three constants:
+BIOS, UEFI and SecureBoot UEFI. The first two are obvious causing selection of
+the correct DHCP filename when deploying TFTP configuration, the third one
+select "shim.efi" chain loading capability for secure booting.
+
```
ohadlevy: ideally i want to avoid defining per hostgroup, maybe we need to introduce a provisioning profile (os, arch, image / boot mode, model(?) )

text/0000-PXE-Booting-UEFI.md
```diff
@@ -0,0 +1,228 @@
...
+Similarly to root password, a global setting could be provided for one or both
+flags for missing values (open item).
+
+### [Foreman installer #14920](http://projects.theforeman.org/issues/14920) - Change dhcpd.conf to support EFI loaders
+
+Default DHCPD configuration must be changed in order to allow booting unknown
+UEFI hosts as it was shown above (dhcpd.conf snippet). Using shim.efi is
+perhaps better default than loading Grub directly.
+
```
ohadlevy: any reason to not default to it on all cases?

text/0000-PXE-Booting-UEFI.md
```diff
@@ -0,0 +1,228 @@
...
+It looks like Grub2 is the best and most compatible way of PXE booting EFI
+hardware and it is available in most Linux distributions. It makes sense to
+provide support in our Puppet classes and installer to deploy these.
+Instructions on where to find the firmware in RHEL7 are at
+[access.redhat.com](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Installation_Guide/chap-installation-server-setup.html)
+(RHEL6 does not ship Grub2 so installer should fail or skip that). Similarly
+for Debian and Ubuntu it is available for [direct
+download](http://archive.ubuntu.com/ubuntu/dists/trusty/main/uefi/grub2-amd64/current/)
+or in debian-installer-8-netboot-amd64 package.
```
ohadlevy: meaning it will work out of the box only for rhel7? or do you expect we carry this version of grub in our repos?

text/0000-PXE-Booting-UEFI.md
```diff
@@ -0,0 +1,228 @@
...
+
+UEFI shim shall be installed if possible. Expected file paths are defined in
```
ohadlevy: any reason why it should not be possible?

text/0000-PXE-Booting-UEFI.md
```diff
@@ -0,0 +1,228 @@
...
+could pass this information when creating new VMs so they gets created with
+correct firmware and PXE booting will just work.
+
+### [Foreman discovery #14922](http://projects.theforeman.org/issues/14922) - Discovery can detect BIOS/EFI and report it back
+
+Discovery can detect current OS loader via a [custom
+fact](https://github.com/jcpunk/puppet-efi/blob/master/lib/facter/efi.rb) and
+this can be used for provisioning to set the "loader" host attribute. Detection
+of secure boot should be possible too, it will be more challenging (depends on
+distribution).
```
ohadlevy: but distribution is only discovery image isn't it? or you mean existing hosts?

text/0000-PXE-Booting-UEFI.md
```diff
@@ -0,0 +1,228 @@
...
+
+* UEFI for ARM/ARM64 platform is emerging and it looks like Grub2 will support it - we should test this
+* PPC/PPC64 uses yaboot or petitboot and it should be relatively easy to add support for this but we need someone to test this
+* IA64 uses eLilo - ditto
+* z360 can be booted via zPXE and Cobbler supports that - can we find someone from the community to help with testing that?
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+* Shall we drop OS variant in favor of pxe template kind?
```
ohadlevy, line 135: what do you mean by os variant? drop what exactly?

text/0000-PXE-Booting-UEFI.md
lzap: > can't we extend the model object instead? would set the default boot flag on the hardware model.

I considered that, if you mean Hardware Model. It has two issues. First, when provisioning with Foreman you usually don't set it. Second, is it's being set by Puppet during initial puppet upload and we have the same problem as with interfaces - need of introducing global setting that will turn off or on this behavior globally. Having it part of host/hostgroup seems cleaner approach to me.

> another option, is that we deploy all kind of configuration files in one action (e.g. we could setup pxelinux and grub at the same time regardless of how the user uses it?)

We could and that was my initial idea, but from what I've learned DHCP "filename" option only accepts one entry and if it fails, then there is no way back when it fails. And conditionals are not possible. We need to know exactly what the client is capable of booting if we want to do reservation.

Edit: I mean, it could be possible to do this with ISC DHCP (when filename is missing from a reservation default is taken which can be conditional), but how about with other DHCP servers? We can perhaps have some special value "Default" or "Per platform" that will only work with ISC DHCP?

text/0000-PXE-Booting-UEFI.md
lzap: Hmm I see it as another option next to os and arch, but if you insist, I can introduce that kind of thing. But IMHO another profile will make it harder for newcomers. It is already quite complicated to get things provisioning.

text/0000-PXE-Booting-UEFI.md
lzap: Not sure if I follow? You want to be Secure Boot to be enabled for all unknown and known hosts?

text/0000-PXE-Booting-UEFI.md
lzap: No, what I mean by that is that for RHEL/CentOS we should use the official Grub1/Grub2/Shim triple which is distributed in RHEL repositories. For all other distributions, we can use Grub2 (Signed or Unsigned version) from Ubuntu as they provide direct download link. This should be good starting point.

text/0000-PXE-Booting-UEFI.md
lzap: Not all distributions uses Shim approach, for example Debian family signs Grub itself and they distribute two copies of the same. https://wiki.ubuntu.com/UEFI/PXE-netboot-install

Shim is not available at all: https://packages.debian.org/search?keywords=shim&searchon=names&suite=stable&section=all

text/0000-PXE-Booting-UEFI.md
lzap: Yeah, scratch that. I only need to investigate how to determine secure boot on Fedora/CentOS.

text/0000-PXE-Booting-UEFI.md
lzap, line 135: To me, operatingsystem@pxe_variant seems like unuseful thing - we only use it with 1:1 mapping with PXE templates:
- syslinux with PXELinux
- pxegrub with PXEGrub
- poap with POAP
- ztp with ZTP

We can simply get rid of pxe_variant, we know the variant thank to the PXE template associated with the OS.

lzap:
This PR was created by Greg, let's just continue the discussion @ohadlevy here and once we solve all your comments, I will file another PR with updated text.

text/0000-PXE-Booting-UEFI.md
timogoebel: > can't we extend the model object instead?

VMWare allows you to boot either via PXE or EFI. PXE is default, but EFI is required for IPv6. The model is "VMWare Virtual Platform" either way. When the setting is stored in the host model, you could set up the compute resource accordingly.

text/0000-PXE-Booting-UEFI.md
ohadlevy: i was just referring to: _perhaps better default than loading Grub directly_ - so i was reading it as enable shim by default - if it makes any sense?

text/0000-PXE-Booting-UEFI.md
ohadlevy: my point is that it seems like its going to be settings that are  not unique per host, rather per a group of hosts, so its either the hw model, or a new join table (e.g. os profile) that should accept it as a paramter.
I know its already overloaded, hence I'm hoping an OS profile (or maybe some other object) is a better name to represent that instead of keep on enlarging host/groups. after all, we are working hard to reduce the size of the host object using facets with @shimshtein

lzap: > VMWare allows you to boot either via PXE or EFI. PXE is default, but EFI is required for IPv6. The model is "VMWare Virtual Platform" either way. When the setting is stored in the host model, you could set up the compute resource accordingly.

Sure, the same works if it's stored for each individual host record.

## 

Later,
 Lukas #lzap Zapletal

text/0000-PXE-Booting-UEFI.md
lzap: > i was just referring to: _perhaps better default than loading Grub directly_ - so i was reading it as enable shim by default - if it makes any sense?

Yeah, we will setup shim when appropriate, sure.

## 

Later,
 Lukas #lzap Zapletal

text/0000-PXE-Booting-UEFI.md
lzap: > my point is that it seems like its going to be settings that are  not unique per host, rather per a group of hosts, so its either the hw model, or a new join table (e.g. os profile) that should accept it as a paramter.
> I know its already overloaded, hence I'm hoping an OS profile (or maybe some other object) is a better name to represent that instead of keep on enlarging host/groups. after all, we are working hard to reduce the size of the host object using facets with @shimshtein

I can introduce provision profile with the following fields:
- architecture
- OS
- ptable
- media

The thing is I would like to hit the 1.12 timeframe with this. This is a
huge refactoring. In this light, attaching it to the hardware model
seems like a better idea, we can add similar global setting to avoid
puppet fact to override the setting.

Others have opinions?

## 

Later,
 Lukas #lzap Zapletal

ohadlevy: @lzap I agree, this is outside of the scope of this RFC - thanks.

ShimShtein: Disclaimer: this is off-topic, but right now it's more of a vague idea so it still doesn't deserve its own RFC.
This discussion got me thinking about the provisioning process:

Can we look at it as a set of provisioning _scenarios_? Each scenario will have a different set of properties that need to be set, and these sets don't have to be similar to each other. For each host we can decide what is the best scenario for it, and once we decide it, we will ask the user/use some kind of defaults to populate the relevant properties. For example provisioning from an image will require only the image to be set, but provisioning from network will need the operating system, architecture, media and so on. Even further, we can create some kind of scenario builder that will adapt the property set accordingly, like in the case @lzap is describing: basically the network provisioning scenario will ask the user whether he wants to provision using grub or pxe. Once the configuration is done, the scenario will start changing all relevant parts like DHCP records according to the choices the user did.

My thought is that each host can be provisioned to the same state using different scenarios.

If you think that the basic idea worth something - tell me, I will think of it more and create an actual RFC for it.

lzap: > Can we look at it as a set of provisioning _scenarios_? Each scenario will have a different set of properties that need to be set, and these sets don't have to be similar to each other. For each host we can decide what is the best scenario for it, and once we decide it, we will ask the user/use some kind of defaults to populate the relevant properties. For example provisioning from an image will require only the image to be set, but provisioning from network will need the operating system, architecture, media and so on. Even further, we can create some kind of scenario builder that will adapt the property set accordingly, like in the case @lzap is describing: basically the network provisioning scenario will ask the user whether he wants to provision using grub or pxe. Once the configuration is done, the scenario will start changing all relevant parts like DHCP records according to the choices the user did.

I am not sure what is the added value of the scenario. Image-based
provisioning still needs an architecture and OS to be set. I think the
more accurate place for BIOS/EFI switch is Hardware Model. Pulling
provisioning info into separate model indeed simplifies host/hostgroup,
but on the other hand it kills the flexibility which is enormous at the
moment - you can decide to select different OS today anytime you want.

I have to be missing something in your explanation.

## 

Later,
 Lukas #lzap Zapletal

GregSutcliffe:
@lzap probably best to keep the discussion in one place, just send a patch to my branch and I'll update the PR - sorry for the mess, was trying to save you some work ;)

lzap:
Extra note for myself: we should add also support "None" setting for pxe_template flag for case iPXE is used with a hypervisor that supports it directly (e.g. libvirt, oVirt). In this case, the VM does not PXE boot at all, iPXE firmware is able to directly load iPXE script from Foreman.

Currently, users must associate a "fake" PXELinux template in order to be able to enter Build mode.

text/0000-PXE-Booting-UEFI.md
lzap: Removing.

text/0000-PXE-Booting-UEFI.md
lzap: Sure, will remove.

text/0000-PXE-Booting-UEFI.md
lzap: Github is not very good in tables, you need to use horizontal scrollbar to see the last column, can't fix as the table has lot of data in it.

text/0000-PXE-Booting-UEFI.md
lzap: I will go ahead with Hardware Model flags. We can create additional ticket to avoid misbehavior for VMWare as @timogoebel pointed out - Hardware Model can be imported from both model and BIOS/EFI (custom defined) fact:

```
Hardware model fact importing will be enhanced - it will be possible to import
the loader flag from an extra custom fact (defined via global settings).
```

http://projects.theforeman.org/issues/12634

AFAIK facter does not include BIOS/EFI kind of fact, but it's easy to add one for Linux.

lzap:
Force pushed, please re-review. If there are no objections by EOB let's merge this and start working.

evgeni:
two things:
- the RFC speaks about EFI and UEFI in different places, I think those two are different and we should only use the UEFI name, no?
- sadly there are systems that have a x86_64 CPU but only 32-bit UEFI (thanks Intel), do we care for those? see https://blogs.intel.com/evangelists/2015/07/22/why-cheap-systems-run-32-bit-uefi-on-x64-systems/ https://tails.boum.org/blueprint/UEFI/32-bit/ https://bugs.launchpad.net/ubuntu/+source/grub2/+bug/1341944 https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=775202

GregSutcliffe:
> Force pushed, please re-review. If there are no objections by EOB let's merge this and start working.

Commits should not be squashed, as per the guidelines in the README. Sqashing commits means we lose the context of earlier comments, as well as the history of things that were considered but dropped. If you still have the reflogs, it'd be worth restoring the state of the branch.

lzap:
> - the RFC speaks about EFI and UEFI in different places, I think those two are different and we should only use the UEFI name, no?

Yeah, changed all to UEFI.

> - sadly there are systems that have a x86_64 CPU but only 32-bit UEFI (thanks Intel), do we care for those? see https://blogs.intel.com/evangelists/2015/07/22/why-cheap-systems-run-32-bit-uefi-on-x64-systems/ https://tails.boum.org/blueprint/UEFI/32-bit/ https://bugs.launchpad.net/ubuntu/+source/grub2/+bug/1341944 https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=775202

We will take advantage of shim loader for Red Hats as it can detect and
load appropriate version:

https://github.com/rhinstaller/shim

Debian based distros don't use shim and also don't provide 32bit
loaders, but users can still use shim as well in this case:

http://archive.ubuntu.com/ubuntu/dists/trusty/main/uefi/grub2-amd64/current/

For SecureBoot, LinuxFoundation's PreLoader is an alternative option for
non-signed distributions like Debian:

http://www.rodsbooks.com/efi-bootloaders/secureboot.html#preloader

## 

Later,
 Lukas #lzap Zapletal

lzap:
> Commits should not be squashed, as per the guidelines in the README. Sqashing commits means we lose the context of earlier comments, as well as the history of things that were considered but dropped. If you still have the reflogs, it'd be worth restoring the state of the branch.

Given amount of small changes, I am keeping it as is, but I pushed
additional change already.

## 

Later,
 Lukas #lzap Zapletal

lzap:
Thanks for comments, I will start with implementation.

ohadlevy:
thanks, please update the rfc readme accordingly.

lzap:
Done.

GregSutcliffe:
You forgot to update the RFC number (no, we are numbering from 0, it'll cause conflicts with the template sooner or later) - I'll fix it ;)

PR #3, opened 2016-05-10
Update asset management

tbrisker:


text/0000-update_asset_packaging.md
```diff
@@ -0,0 +1,101 @@
...
+* Lack of examples in other projects
+
+__Option 3: Browserify + npm__
+
+[POC PR](https://github.com/theforeman/foreman/pull/3326)
+
+_Pros:_
+* Simple installation and configuration.
+* Easy to tie into rails asset pipeline.
+* Easy to add es2015, minification, etc.
```
tbrisker: Wouldn't this require adding another task runner?

tstrachota: That imho depends on whether we use `browserify-rails` which should take care of rebuilding the assets, or plain `browserify` which would need to be executed via gulp or similar.

tbrisker: would that also work for our own code? e.g. if we want to start using es2015 in Foreman core js?

text/0000-update_asset_packaging.md
```diff
@@ -0,0 +1,102 @@
...
+
+_Pros:_
+* Bower seems to have a larger ecosystem for front-end assets then npm
+* Bower is the current standard in use across web applications
+* Already in use in Katello, amount of effort required is halved.
+* Live reload development server could be done with some more work to grunt/gulp or any other build system
+* Bower uses flat dependency tree to reduce dependency duplicities that clients would have to download if plain npm + browserify was used.
+
+_Cons:_
+* Bower project future is uncertain - currently one maintainer working in his spare time.
```
waldenraines: I am not sure this is still the case, can we provide evidence for this statement?

text/0000-update_asset_packaging.md
```diff
@@ -0,0 +1,102 @@
...
+* Adds two more package managers instead of one - requires npm to install bower. This means we will have 3 different places that contain dependencies.
+* Some packages now are packaged only in npm
```
waldenraines: Do you have any examples of web dependencies that are only in npm?

text/0000-update_asset_packaging.md
```diff
@@ -0,0 +1,102 @@
...
+* Does not provide solution for any server-side JS (testing, transpiling, etc.). Any such dependencies will need to be handled either by npm or bundler.
```
waldenraines: This is a purposeful separation of concerns. Bower is concerned only with client side web related things.  Therefore you can expect to see jQuery in bower but not http-server.  Do we expect to be running any server-side JS in foreman/katello?

text/0000-update_asset_packaging.md
```diff
@@ -0,0 +1,102 @@
...
+
+See above for reasons for and against.
+
+__Option 2: Webpack + npm__
+
+[POC PR](https://github.com/theforeman/foreman/pull/3433)
+
+_Pros:_
+* Live reload development server.
+* Ability to package the code in separate chunks that are only loaded when needed.
```
waldenraines: This is the case in production as well or does webpack concatenate and minify everything together into a small number of files?

text/0000-update_asset_packaging.md
```diff
@@ -0,0 +1,102 @@
...
+* Multiple loaders allow for server-side tasks such as transpiling, linting, minifing etc. to be run without requiring an additional task runner.
+* Can possibly replace rails asset pipeline in the future if desired.
```
waldenraines: I don't think this pro is limited to webpack so it should either be removed or added to all other options.

text/0000-update_asset_packaging.md
```diff
@@ -0,0 +1,102 @@
...
+_Cons:_
+* Only handles JS assets.
+
+__Option 4: Bower__
+
+[POC is part of PR](https://github.com/theforeman/foreman/pull/2882)
+
+_Pros:_
+* Bower seems to have a larger ecosystem for front-end assets then npm
+* Bower is the current standard in use across web applications
```
waldenraines: I added this line but it's really a duplicate of the line before it so we can remove it.

text/0000-update_asset_packaging.md
```diff
@@ -0,0 +1,102 @@
...
+* Already in use in Katello, amount of effort required is halved.
+* Live reload development server could be done with some more work to grunt/gulp or any other build system
+* Bower uses flat dependency tree to reduce dependency duplicities that clients would have to download if plain npm + browserify was used.
+
+_Cons:_
+* Bower project future is uncertain - currently one maintainer working in his spare time.
+* Adds two more package managers instead of one - requires npm to install bower. This means we will have 3 different places that contain dependencies.
```
waldenraines: Four if you count RPMs 😉

waldenraines:
@tbrisker thank you for turning this into an RFC 👍

text/0000-update_asset_packaging.md
ohadlevy: The webpack option with live reload requires a web server (with websockets support for live reload). so in dev environment this could be beneficial.

text/0000-update_asset_packaging.md
ohadlevy: also in production. (there is little difference between pro/dev in webpack, as you also have map files).

text/0000-update_asset_packaging.md
dLobatog: I can provide (sort of) evidence for the opposite - http://bower.io/blog/2015/bower-is-growing-join-us/

text/0000-update_asset_packaging.md
dLobatog: 'web dependencies' is the key here. I think npm has 73k packages vs 13k of bower, but then again npm audience expects stuff like yeoman, grunt, etc.. to be installed via 'npm install' while bower is supposed to be only for your front-end assets.

text/0000-update_asset_packaging.md
```diff
@@ -0,0 +1,102 @@
...
+* Some packages now are packaged only in npm
+* Does not provide solution for any server-side JS (testing, transpiling, etc.). Any such dependencies will need to be handled either by npm or bundler.
+
```
dLobatog: Another con of bower IMO is the fact dependencies are loaded in the global space. This means they can be accidentally accessed and changed. Which is what we have already with app/assets/javascripts though.

npm uses CommonJS modules which is basically namespacing your JS, which means you just won't accidentally mess with your dependencies.

text/0000-update_asset_packaging.md
```diff
@@ -0,0 +1,102 @@
...
+__Option 1: Stay with current solution__
+
+See above for reasons for and against.
+
+__Option 2: Webpack + npm__
+
+[POC PR](https://github.com/theforeman/foreman/pull/3433)
+
+_Pros:_
+* Live reload development server.
```
dLobatog: Another pro of npm - https://github.com/theforeman/rfcs/pull/3/files#r62695789

text/0000-update_asset_packaging.md
```diff
@@ -0,0 +1,102 @@
...
+* Complex configuration compared to other alternatives.
+* Lack of examples in other projects
+
+__Option 3: Browserify + npm__
+
+[POC PR](https://github.com/theforeman/foreman/pull/3326)
+
+_Pros:_
+* Simple installation and configuration.
+* Easy to tie into rails asset pipeline.
```
dLobatog: Another pro of npm - https://github.com/theforeman/rfcs/pull/3/files#r62695789

text/0000-update_asset_packaging.md
dLobatog: Another con of bower: hard to concatenate and minify all your assets

text/0000-update_asset_packaging.md
waldenraines: Right, so how is this con different from the one above?

text/0000-update_asset_packaging.md
waldenraines: Correct and that's how grunt works as well.  I should have specified non-dev-tool-server-side-JS.  I think npm will be used for server side dev tools in either case.

text/0000-update_asset_packaging.md
waldenraines: > Another con of bower IMO is the fact dependencies are loaded in the global space. This means they can be accidentally accessed and changed. Which is what we have already with app/assets/javascripts though.

When you say "loaded in the global space" are you referring to the JS runtime environment or where the actual files themselves end up on disk?  If the latter then that is easily configurable. 

> Another con of bower: hard to concatenate and minify all your assets

I think a lot of these "cons" come from a misunderstanding of bower's purpose.  Bower is for dependency management and nothing else.  To concatenate and minify you would use a build tool such as grunt or gulp and both of these tasks are trivial with either build tool.

text/0000-update_asset_packaging.md
waldenraines: To me this isn't a pro, it's a con.

The reason being is that browsers can only make so many simultaneous connections to a host and that once they download something it is cached locally.  I would imagine that a single concatenated, minified, and gzipped JS file that contained all of foreman's JS code downloaded once would make the application more responsive than forcing the browser to download new files on every page.  Granted that initial page load will take a little longer but after that you're golden.

ETA: Please excuse my ignorance of webpack.  I haven't had enough time to look into it as much as I would like.

dLobatog: I'm of the same opinion of @waldenraines here at least intuitively. When I did tests for react, it was significantly slower to not concatenate, minify, etc... everything properly. I'll try to report data when I have it for webpack though

ehelms: How big are the webpack chunks? Is that configurable? Having 3-5 chunks might be quicker than having a single massive file. And when I say quicker, I mean relative to the user's experience visiting a page. This could also be a general design concern for all pages and plugins -- having "common" assets in one chunk and loading specific chunks depending on sections of the application similar to what we kind of do now with plugins providing their own assets.

text/0000-update_asset_packaging.md
ehelms: As well, which might be good to quote for others if we are worried about the maintenance of the project, is the pulse page -- https://github.com/bower/bower/pulse/monthly

text/0000-update_asset_packaging.md
ehelms: I think you can use https://github.com/rharriso/bower-rails to avoid npm install and then bower install if it bothers us. In Bastion, we use npm and grunt to handle unit tests, bower, linting, etc. So it really only boils down to NPM and a task runner.

text/0000-update_asset_packaging.md
```diff
@@ -0,0 +1,102 @@
...
+* Only one new tool to add to the project.
+
+_Cons:_
+* Requires means of adding the npm modules to the assets pipeline, such as a task runner or rake task.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+Issues that are common to any alternative chosen:
+* How to package rpms downstream w/o internet connection?
```
ehelms: This is not a downstream concern, this is an upstream concern first and foremost as Koji does not allow internet connection.

text/0000-update_asset_packaging.md
waldenraines: > Having 3-5 chunks might be quicker than having a single massive file. And when I say quicker, I mean relative to the user's experience visiting a page.

On the initial page sure, but after that initial load having multiple assets that differ on every page will be slower because the server will have to go acquire those files whereas the single file would be cached.  Subsequent page loads would be much faster giving the application a "snappy" feel.

text/0000-update_asset_packaging.md
tbrisker: Very well, removing this line. It originated from the crisis the project underwent several months ago, glad to hear they bounced back.

text/0000-update_asset_packaging.md
tbrisker: Currently, almost all of the JS code in foreman core is executed on every page load, including a large amount that is irrelevant on most pages. It also leads sometimes to unexpected side-effects, when code aimed with a certain page in mind is executed on other pages.
I don't think that the time saved by downloading everything in one big chunk is that significant, compared to the benefits of splitting the code into modules that are only loaded when needed.

text/0000-update_asset_packaging.md
tbrisker: Is it? I'm not sure that all the other options allow for a complete replacement including cache-busting, minification, etc.

text/0000-update_asset_packaging.md
tbrisker: removing

text/0000-update_asset_packaging.md
tbrisker: Looks like it would still require npm: https://github.com/rharriso/bower-rails#bower-installation

text/0000-update_asset_packaging.md
tbrisker: corrected.

tbrisker:
Thanks everyone for your comments, updated the RFC with the agreed points.

text/0000-update_asset_packaging.md
```diff
@@ -0,0 +1,103 @@
...
+
+__Option 4: Bower__
+
+[POC is part of PR](https://github.com/theforeman/foreman/pull/2882)
+
+_Pros:_
+* Bower seems to have a larger ecosystem for front-end assets then npm
+* Already in use in Katello, amount of effort required is halved.
+* Live reload development server could be done with some more work to grunt/gulp or any other build system
+* Bower uses flat dependency tree to reduce dependency duplicities that clients would have to download if plain npm + browserify was used.
```
tbrisker: Seems like NPM also does that starting with version 3: https://docs.npmjs.com/how-npm-works/npm3

tbrisker:
Looks like pushing the updated version caused some comments to be incorrectly hidden, please note the comments on outdated diffs as well.

text/0000-update_asset_packaging.md
tstrachota: And they also provide [npm dedupe](https://docs.npmjs.com/cli/dedupe) tool for flattening the tree. I tested that today and it seems to work fine.

text/0000-update_asset_packaging.md
```diff
@@ -0,0 +1,103 @@
...
+* npm uses CommonJS modules which is allows namespacing JS code, which means you just won't accidentally mess with your dependencies.
+
+_Cons:_
+* Requires means of adding the npm modules to the assets pipeline, such as a task runner or rake task.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+Issues that are common to any alternative chosen:
+* How to package rpms w/o internet connection, as Koji does not allow internet connection?
```
tstrachota: Don't we do something similar for puppet modules in installer packaging already?

text/0000-update_asset_packaging.md
tstrachota: I haven't tried that personally but it should work https://github.com/browserify-rails/browserify-rails#using-browserify-transforms

text/0000-update_asset_packaging.md
tstrachota: > Another con of bower: hard to concatenate and minify all your assets 

I don't think it's harder than with any other tool. It's just matter of setting correct paths in your minifier/runner.

tstrachota:
I think we're comparing selected combinations of tools that can be split by their purpose into following groups:

**A) package managers**
- [bundler](http://bundler.io/)
- [npm](https://www.npmjs.com/)
- [bower](http://bower.io/)

**B) asset processing**
- **Ba) task runners**
  - [grunt](http://gruntjs.com/)
  - [gulp](http://gulpjs.com/)
- **Bb) package bundlers**
  - [browserify](http://browserify.org/)
  - [webpack](https://webpack.github.io/)
  - [rails asset pipeline](http://guides.rubyonrails.org/asset_pipeline.html)

In fact any combination of A x B (or A x Ba x Bb) is possible (currently we're on bundler + asset pipeline). It seems to me that we're often repeating pros and cons in the explored alternatives, because they are actually pros/cons of the tools from the list above. I think it would make sense to compare tools from the same category and then select the combination of those that we find best.

text/0000-update_asset_packaging.md
ehelms: We generate a source tarball that pulls in the assets, that tarball is then used to do the RPM build. The two processes are separate but pipelined in Jenkins.

text/0000-update_asset_packaging.md
waldenraines: > Currently, almost all of the JS code in foreman core is executed on every page load, including a large amount that is irrelevant on most pages. It also leads sometimes to unexpected side-effects, when code aimed with a certain page in mind is executed on other pages.

This is due to a lack of namespacing and a lack of using a framework more than anything else.  In Foreman everything is defined on globally on `window` whereas in a modern webapp things are modularized.  We never have the side-effects between pages that you mention here in katello because everything is properly namespaced (even though all of the JS is delivered in a single file in production mode).

And actually, come to think of it, katello in production mode vs development mode is a good example of how much faster a single concatenated, minified, and gzipped file can be than a bunch of other files.

> I don't think that the time saved by downloading everything in one big chunk is that significant, compared to the benefits of splitting the code into modules that are only loaded when needed.

Then why do most sites concatenate all of their files into one?  Why don't we see a trend of modularization by file instead of by namespacing?

text/0000-update_asset_packaging.md
waldenraines: You can do a live reload development server with any of these solutions and thus it should be removed as a pro from webpack.

text/0000-update_asset_packaging.md
```diff
@@ -0,0 +1,109 @@
...
+* Single package management system - bundler. All dependencies are in one central location, build process is simple.
+* Rails asset pipeline doesn’t require any modifications to handle the current solution.
+
+# Alternatives
+[alternatives]: #alternatives
+
+__Option 1: Use Grunt task runner__
+
+_Pros:_
+* Existing knowledge base and tasks from Katello/Bastion.
```
waldenraines: Pro for both grunt and gulp is that there is already a large amount of plugins available that do common tasks so we won't have to write them ourselves.  This  may be the case with webpack as well, I'm not sure.

text/0000-update_asset_packaging.md
```diff
@@ -38,66 +41,69 @@ Some reasons to stay with the current solution:
...
 * Ability to package the code in separate chunks that are only loaded when needed.
-* Multiple loaders allow for server-side tasks such as transpiling, linting, minifing etc. to be run without requiring an additional task runner.
+* Multiple loaders allow multiple tasks be run without requiring an additional task runner and with minimal configuration.
 * Can possibly replace rails asset pipeline in the future if desired.
-* npm uses CommonJS modules which is allows namespacing JS code, which means you just won't accidentally mess with your dependencies.
 
 _Cons:_
-* Complex configuration compared to other alternatives.
-* Lack of examples in other projects
+* More complex configuration compared to webpack.
```
gailsteiger: Webpack con: more complex ... compared to webpack? Recursive?

tbrisker: oops, that was supposed to be compared to browserify

text/0000-update_asset_packaging.md
tbrisker: webpack allows this as part of its core functionality and it is integrated in its build process, while others would require setting up the server separately if I understand correctly (but I may be unaware of some similar solution for others).

text/0000-update_asset_packaging.md
```diff
@@ -38,66 +41,69 @@ Some reasons to stay with the current solution:
...
-* Bower uses flat dependency tree to reduce dependency duplicities that clients would have to download if plain npm + browserify was used.
-
-_Cons:_
-* Adds two more package managers instead of one - requires npm to install bower. This means we will have 3 different places that contain dependencies.
-* Some packages now are packaged only in npm
-* Does not provide solution for any server-side JS (testing, transpiling, etc.). Any such dependencies will need to be handled either by npm or bundler.
-
-__Option 5: npm alone__
-
-No POC available.
```
waldenraines: Should we keep this in here for historical reasons?  Otherwise, once you squash these commits no one is going to know we considered bower.

tbrisker: If I understand the RFC process correctly then we aren't supposed to squash commits precisely for that reason.

waldenraines: Ah, okay 👍

text/0000-update_asset_packaging.md
tbrisker: > Then why do most sites concatenate all of their files into one? Why don't we see a trend of modularization by file instead of by namespacing?

Many websites don't in fact concatenate everything into one file, but rather a few. The basics are all usually in a couple of files that are loaded on all pages with extras for certain pages. For example, twitter has a different js file loaded on the profile page, as does stack overflow. I agree we should make the code more modular and namespaced, but in some cases it may make sense have a certain portion of the code load only for some of the pages and not others.

Another thing we should keep in mind is that our use-case is different then most major websites - a large website should be optimized to handle thousands and even miliions of connections per second, while even the heaviest of our users will only need to handle several dozen per second at most. In addition, most of our users are connected via LAN to the server, rather then via a lagging internet connection. 
In almost all of our pages, calculating the right data from the database (or querying external sources for it) takes significantly more time then actually sending the files to the users.
I don't think we should have 20 different js files for every page - but also going to the other extreme of always serving everything in one file is not the right way in my opinion.

text/0000-update_asset_packaging.md
```diff
@@ -0,0 +1,109 @@
...
+__Option 2: Use Gulp task runner__
+
+_Pros:_
+* Runs faster then Grunt and with less configuration.
+* Tasks can be built as streams of smaller units.
+
+_Cons:_
+* Each task needs manual set up and configuration.
+* Little relevant knowledge within the current core team.
+
```
waldenraines: I'm not sure why we're adding grunt and gulp are in here since you can use webpack with [grunt](http://webpack.github.io/docs/usage-with-grunt.html) and [gulp](http://webpack.github.io/docs/usage-with-gulp.html).  To me, webpack and grunt/gulp accomplish separate things.

waldenraines:
I am more than a little confused about what we are choosing here.

I agree with the categories defined in [@tstrachota's comment](https://github.com/theforeman/rfcs/pull/3#issuecomment-219411736) and I think we should be looking at these categories independently.  

In the last meeting, for example, we compared a package manager (bower) to two module bundlers (webpack and browserify) and chose the two module bundlers even though both [webpack](http://webpack.github.io/docs/usage-with-bower.html) and [browserify](https://www.npmjs.com/package/browserify-bower) can be used with bower (i.e. they accomplish separate things).  This is just like choosing webpack over npm, it makes no sense.  

I think any vote should be drawn around these lines (mostly the same as what @tstrachota proposed above):

**Package Managers**
- bundler
- npm
- bower

**Task Runners**
- grunt
- gulp

**Module Bundlers**
- Browserify
- Webpack
- Rails asset pipeline
- None

**Javascript Frameworks**
- AngularJS
- Ember
- Etc.

So what is the purpose for this RFC, which category is this RFC for?

text/0000-update_asset_packaging.md
waldenraines: > Many websites don't in fact concatenate everything into one file, but rather a few.

That's true, I'm not sure if there is a consensus on best practices here.  I would imagine that it has to do with the size of your file.

> I don't think we should have 20 different js files for every page - but also going to the other extreme of always serving everything in one file is not the right way in my opinion.

After thinking about this some more I tend to agree.  I think having something like one file for libraries, one for our shared code, and then one per abstract UI components could be okay.  What I think we should avoid is having one per page or per specific UI component (i.e. hosts.js, smart_proxies.js, host-table.js, etc).   Of course I am talking about production mode only, in development such distinctions may make sense.

text/0000-update_asset_packaging.md
```diff
@@ -0,0 +1,57 @@
...
+After a certain period of attempting this solution, we will review the choice and decide if we wish to continue in this direction or change to a different one.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Some reasons to stay with the current solution:
+* No extra work needed.
+* Single package management system - bundler. All dependencies are in one central location, build process is simple.
+* Rails asset pipeline doesn’t require any modifications to handle the current solution.
+
```
ohadlevy: I would add that it would be potentially harder to use the theme  plugins. @ShimShtein ?

text/0000-update_asset_packaging.md
```diff
@@ -0,0 +1,57 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+Several other alternatives were considered before choosing this option. The alternatives considered:
+
+For package management - NPM, Bower, remain with current solution.
+
+For handling the assets (getting them working with foreman) - Grunt, Gulp, Webpack, Browserify, Rake.
```
ohadlevy: can we close the other POC prs? maybe link them here too?

text/0000-update_asset_packaging.md
ShimShtein: Lt's examine theme plugin extension points:
1. Images: are added by modifying Rack's pipeline to serve theme's static files before serving those from the main app.
2. css: Should contain only small amounts of css rules that are necessary for adapting to the theme - they are linked at the bottom of the head element (by using deface) and are served by theme's engine.
3. Javascripts: Should contain only overrides that are relevant to the theme - they are linked at the bottom of the head element  too. The only drawback that I can think of, is introduction of modules and namespaces in the Javascript, making it more difficult to find the right method to override.

All in all, this design should not interfere with themes design, except the issue with namespaced javascript.
@tbrisker, what do you have in mind regarding this kind of issues? Do you plan adding namespaced approach to JS and if so, how would you recommend redefining methods from the main app.
Anyway I would like to see a POC that contains a theme before we can say that it's a viable option.

tbrisker: @ShimShtein could you please test this POC with the theme plugin and see if this causes any issues? I'm not sure what kind of POC with theme precisely you expect.
I would also say that it would make more sense to change the theme plugin if needed to work with this rather then block this due to the theme plugin. We shouldn't hold off taking the core forward because a plugin would need adjustments to play well with the changes.

ShimShtein: I don't care making adjustments, as long as they are _possible_. I am not sure that the theme is the only plugin that overrides JS behavior, it's just an example for a heavy UI changing plugin.
Foreman is behaving as a platform for this matter, so we do need to think about plugin compatibility. IMHO, it would be fare to provide a migration path, if we change some behavior.

@tbrisker From my point of view, the only non-trivial concern is javascript overrides, can you tell if there will be changes to JS object model?
Anyway I will gladly test the theme against this POC.

text/0000-update_asset_packaging.md
dLobatog: Every POC but the webpack PR with live reload is closed now

text/0000-update_asset_packaging.md
```diff
@@ -0,0 +1,57 @@
...
+The node modules will be integrated into foreman using Webpack. Webpack has a powerful system of loaders that allow handling various tasks. Webpack also comes with a small webserver that allows live-reload of all assets managed by it, hopefully leading to increased developer happiness. In addition, webpack has various other strengths that make it seem like the current leading option - ability to handle css assets and all js module flavors, smart division of code into bundles and more.
+As we will currently continue using the Rails assets pipeline, any plugin will be able to use whatever method they choose to handle any assets as long as the result is added to the pipeline (as is done currently).
+After a certain period of attempting this solution, we will review the choice and decide if we wish to continue in this direction or change to a different one.
```
dLobatog: What are we going to do with the existing assets? I think we should start converting some of them to CommonJS as you started doing on the POC, but I'm not sure if this is feasible for all of them.

tbrisker: Gradually move what we can to being modular. (btw, I used es6 modules here, not commonjs)

text/0000-update_asset_packaging.md
tbrisker: There may be small changes needed in plugins to correctly handle this, but we will need to see and solve each case separately. I don't think there is any sweeping guideline we can provide at this time as use cases very quite a lot (between nearly none to adding complete frameworks such as Angular or Ember).

text/0000-update_asset_packaging.md
```diff
@@ -0,0 +1,61 @@
...
+
+For handling the assets (getting them working with foreman) - Grunt, Gulp, Webpack, Browserify, Rake.
+
+History of the discussion and pros and cons of each option can be read in the previous commits of this file.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+* How to package rpms without internet connection, as Koji does not allow internet connection?
+  * The compiled assets could be committed into source control. This has a downside of committing a large amount of generated code into git, and may lead to issues with forgetting to recompile before commit.
```
ohadlevy: @tbrisker would you update this section to reflect the latest discussions?

tbrisker: @ohadlevy I thought I did - I added all the options of handling that as well as the concern raised by @domcleal about Fedora guidelines, with the intent of commenting on them (and allowing others to do so as well).

text/0000-update_asset_packaging.md
```diff
@@ -0,0 +1,61 @@
...
+  * The `node_modules` directory could be added to source control. This again leads to a large amount of files that aren't required polluting our repository and issues with keeping the directory up to date.
+  * Each node module could be built as a separate package. This is unfeasible, since there are hundreds of modules required for the generation of the minified files. In addition, we would not want to package every single one ourselves, given that one of the goals of this change is to reduce packaging workload.
+  * The `node_modules` directory could be added to the source tarball before being sent to the builder, with the compiling and minification being run on the builder.
```
tbrisker: I have attempted to update the PR to use this approach, which is also used by other projects such as cockpit.

text/0000-update_asset_packaging.md
ohadlevy: right, my assumption is that you write it as the proposal vs unresolved, then if someone disagree he can speak out :)

ohadlevy: ideally with a link to a packaging pr :)

text/0000-update_asset_packaging.md
```diff
@@ -0,0 +1,61 @@
...
+  * The js assets could be compiled and minified before the build and have only the resulting files be added into the tarball passed into the builder.
+  * Some concern has been raised that any option other then packaging every single library may go against the Fedora Packaging Guidelines.
```
tbrisker: I'm not sure how committed we should be to the Fedora guidelines, since:
- There is no plan to distribute Foreman in Fedora in the foreseeable future.
- We already include various js libraries in the `vendor/`directory, not sure why moving some to `node_modules/` would change our compliance.

ohadlevy: my point is to discuss how we are going to package it, not much about
fedora guidelines (which has some effect here), e.g. as part of this RFC we
need an answer to how, and discuss the details and come to an agreement.

text/0000-update_asset_packaging.md
tbrisker: moved this section into the detailed design along with the current proposal for handling it.

text/0000-update_asset_packaging.md
```diff
@@ -0,0 +1,65 @@
...
+* Committing either the `node_modules` directory or the compiled assets could be committed into source control. This has a downside of committing a large amount of generated code into git, and may lead to issues with forgetting to recompile before commit.
+* Compiling and minifying the js assets before the build and add the resulting files into the tarball passed into the builder.
+* Adding the `node_modules` directory to the source tarball before being sent to the builder, with the compiling and minification being run on the builder.
+
+We propose using the last option, as it provides all the needed sources for the builder, in a similar way to how it is currently done for assets included in either gems or under `vendor/` that are needed for the `rake assets:precompile` task to work correctly.
```
tbrisker: @domcleal @ehelms @mmoll would be happy to get your inputs on this, as you are the experts on packaging

tbrisker: Created https://github.com/theforeman/foreman-packaging/pull/1174 and https://github.com/theforeman/foreman-packaging/pull/1175 to handle the packaging side, and https://github.com/theforeman/foreman/pull/3433/files#diff-442dfed27b160ff7165f10d1d0ba9630L26 to handle pushing the `node_modules` dir into the source tarball

text/0000-update_asset_packaging.md
```diff
@@ -26,8 +26,20 @@ Some of the reasons that we want to change the way we currently handle this:
...
+
+Packaging, especially for RPMs, poses a challenge, as Koji builders do not allow internet connection. Building the js assets for production requires running `npm install` to download the needed libraries, followed by `rake webpack:compile` which compiles only the needed portions of the js libraries along with our code into bundles that are minified. Several approaches were considered to handle this:
+
+* Packaging each node module as a separate package. This is unfeasible, since there are hundreds of modules required for the generation of the minified files. In addition, we would not want to package every single one ourselves, given that one of the goals of this change is to reduce packaging workload. However, some concern has been raised that this may be the only approach that fully conforms with the Fedora Packaging Guidelines.
+* Committing either the `node_modules` directory or the compiled assets could be committed into source control. This has a downside of committing a large amount of generated code into git, and may lead to issues with forgetting to recompile before commit.
```
dLobatog: We can enforce that through git hooks (pre-receive or post-receive on github for example)

dLobatog:
Just to note this here - we've agreed on using `-bundle` packages for the package.json dependencies, as per https://github.com/theforeman/foreman-packaging/pull/1174#issuecomment-228024054

ohadlevy:
For me to know... why is that better/different then one package with all
assets in it?

dLobatog:
@ohadlevy Just a way to update individual assets easily instead of having to update the entire assets package, and also fulfill general Fedora recommendations when bundling libraries in a package. More background on https://fedoraproject.org/wiki/Bundled_Libraries?rd=Packaging:Bundled_Libraries &  https://github.com/sfreire/njs2rpm#rationale

dLobatog:
@tbrisker Mind to update this? I'd be happy to merge after - we already implemented & merged this RFC

dLobatog:
@tbrisker ping?

tbrisker:
@dLobatog thanks for the reminder, updated.

text/0003-update_asset_packaging.md
```diff
@@ -0,0 +1,66 @@
...
+* Committing either the `node_modules` directory or the compiled assets could be committed into source control. This has a downside of committing a large amount of generated code into git, and may lead to issues with forgetting to recompile before commit.
+* Compiling and minifying the js assets before the build and add the resulting files into the tarball passed into the builder.
+* Adding the `node_modules` directory to the source tarball before being sent to the builder, with the compiling and minification being run on the builder.
+
+In the end, the approach taken was to package every top-level npm dependancy as a bundle with all its requirements. This reduces the number of npm packages that we need to repackage as rpms from hundredrs of individual modules to only a few dozen, which are all listed in package.json. The process of generating the rpm specs for the node modules is mostly automated, using https://github.com/dLobatog/npm2rpm.
```
dLobatog, line 41: Not all of package.json - only devDependencies needed to run webpack:compile and regular dependencies. :smile:  It's fine but I wanted to clarify

PR #5, opened 2016-06-07
Change design of variant/loader UEFI flags

lzap:
I realized that Hardware model is not part of Hostgroup, therefore it can't be
set for group of hosts which makes it unusable (we don't want users to set
BIOS/UEFI flag for each individual host).

Also since it is possible to deploy all supported template kinds simultaneously, the only flag that is needed is "pxe_loader" which defines the TFTP filename to use. Initially I thought also to setup template kind for each individual host.

The best solutions of storing the "pxe_loader" flag I was able to come up with so far.

1) Keep it associated to Hardware model, introduce Global settings so user can
set it globally and then per individual host. Quite limiting if infrastructure
BIOS vs UEFI is 50/50.

2) Introduce OS profile as it was discussed in
https://github.com/theforeman/rfcs/pull/2#discussion_r62620431 - looks like an
overkill to me. We already have pretty complex setup - template vs operating
system vs default template together with architecture and media. This will make
it worse.

3) Simply add new two fields to the Operating System tab for Hosts and
Hostgroups. As it was mentioned, our ultimate goal is to make the New Host form
more simple, which is not the case.

4) Use Host parameters for the flag. This leverages current design and
workflow giving flexibility of using Hostgroups. The limitation is we don't
support dropdowns in Host parameters, but I assume users will be using
Hostgroups primarily. Hardware model can be extended to provide additional
parameters so we can load up several "predefined" hardware models and users can
still use dropdown per-host basis.

I like the fourth solution the most, objections?

### DIFF
--- a/text/0001-PXE-Booting-UEFI.md
+++ b/text/0001-PXE-Booting-UEFI.md
@@ -6,7 +6,7 @@
 # Summary
 [summary]: #summary
 
-The goal of this feature is to implement PXE booting of UEFI systems. In the default configuration on at least RHEL6, RHEL7 and Debian stable platforms, both BIOS and UEFI systems should be able to PXE boot including Discovery Image. Foreman bootdisk UEFI support is out of scope of this task.
+The goal of this feature is to implement PXE booting of UEFI systems. In the default configuration on at least RHEL6, RHEL7 and Debian stable platforms, both BIOS and UEFI systems should be able to PXE boot including Discovery Image. Foreman Bootdisk UEFI support is out of scope of this task.
 
 # Owners
 
@@ -20,7 +20,7 @@ Currently, Foreman supports booting of x86 BIOS systems via PXELinux and several
 # Detailed design
 [design]: #detailed-design
 
-Initial testing was done to get UEFI working on PXELinux and it looks like stable version of Syslinux (= PXELinux), which is currently 6.02, does not appear to support PXE booting correctly (at least not with TianoCore UEFI ROM in QEMU/KVM). The only version which was found working was 6.03-pre20.
+Initial testing was done to get UEFI working on PXELinux and it looks like stable version of Syslinux (= PXELinux), which is currently 6.02, does not appear to support PXE booting correctly (at least not with TianoCore UEFI ROM in QEMU/KVM). The only version which was found working was 6.03-pre20, but there is no stable release yet with UEFI support.
 
 Additionally, there is no PXELinux version with UEFI support in RHEL6 and RHEL7 and both systems ship with Grub (versions 1 and 2 respectively). Grub is more mature when it comes to UEFI support and since it can boot via PXE and Foreman already supports Grub1, it is more feasible to get UEFI working via officially supported Grub. To have complete support matrix, PXELinux will be part of this implementation, but hardware support (and thus testing) will be limited.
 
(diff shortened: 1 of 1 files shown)


lzap:
I have updated the description and text once again. Opinions?

timogoebel:
I prefer option 3. Let me explain, why: It is very similar to option 4 but has a proper ui and for me as a user doesn't feel "patched in". It does make the host form a bit more complex, but if sane defaults are provided it does not necessarily make it more complicated. And it's still possible to introduce a os profile if we want this at a later point.

timogoebel:
> Introduce OS profile as it was discussed in
> #2 (comment) - looks like an
> overkill to me. We already have pretty complex setup - template vs operating
> system vs default template together with architecture and media. This will make
> it worse.

@lzap: We might have a similar problem with IPv6. How do you tell a host if it should be provisioned via IPv4 or IPv6? I am wondering if a OS profile could help here. We could of course offer both and let the host decide.

lzap:
Are we going to allow two or more provisioning interfaces? I thought there will only be one provisioning one either in IPv4 or IPv6 mode.

GregSutcliffe:
There's no reason we couldn't have more than one provisioning interface (and it could be handy for bonded ethernet, where you don't know which device it will boot from).

The limit of one was chosen in 1.8 during the interfaces redesign, to keep things simple, that's all. The UI might need checking though - I'm not sure how well multiple provisioning interfaces would look with the icons.

timogoebel:
> Are we going to allow two or more provisioning interfaces? I thought there will only be one provisioning one either in IPv4 or IPv6 mode.

Actually I didn't change anything regarding the provisioning interface limitation. Currently only one interface is supported. Let me rephrase: The provisioning interface can have to subnets: the ipv4 subnet and the ipv6 subnet. Both subnets can have a tftp and a dhcp proxy set. Which proxy should we use for provisioning? Both of them?

lzap:
Sorry for the delay, 

> Which proxy should we use for provisioning? Both of them?

Yeah, both. @GregSutcliffe is right about the limitation and this is perhaps the right time to remove it and support multiple orchestrations. This looks cleaner and allows bonded interfaces rather than putting another flag in host or OS profile.

Since this is only one individual flag to add, I am going to leave the OS profile idea out and implement it as a new flag for a host.

lzap:
I've amended the changes and I am starting with coding, let's keep this PR opened for few more days.

lzap:
I just filed https://github.com/theforeman/foreman/pull/3679 with the PXE loader flag implementation which seems to be smootest.

text/0001-PXE-Booting-UEFI.md
```diff
@@ -62,19 +62,24 @@ Foreman Proxy must be enhanced with new PXE variant "Pxegrub2" to allow deployin
 
 ### [Foreman core #12634](http://projects.theforeman.org/issues/12634) - New HW Model flags pxe_template and loader
 
-Hardware model will be enhanced with two new flags: PXE template specifies PXE template kind and variant pair to use. Until today, this was hardcoded to PXELinux or PXEGrub by selected operating system. Optionally, template variant flag from OS could be simply dropped because PXE template implies variant to use and there are no variants which share the same configuration syntax (this is an open item).
+Foreman will support both BIOS and UEFI systems and it is required to associate extra information for each indidivual host (or hostgroup) in order to provide TFTP "filename" option correctly. It can be set per instance via new Host flag (free-form dropdown field) called "PXE Loader" (`pxe_loader` model attribute) with possible values:
 
-Orchestration TFTP code must be modified to use multiple "KIND LOADER default local boot" and "KIND LOADER global default" templates. Local and global boot templates must be added for Grub2 new kind. Naming convention will be:
+* pxelinux.0 (default value)
```
dLobatog: I'm wondering if we should do the following to avoid complicating the life of the user more (I don't think users need to be concerned about which file is used to boot their hosts):
- Start by serving the PXE template with filename  'pxegrub'. If this works,  shortly after we will get a request for the `provision` kind of template. End of the story.
- If it doesn't work, serve the template again with filename 'grub/bootx64.efi'. Rinse and repeat until we find a filename that works.

This assumes the computer will reboot itself after some time if the PXE template wasn't suitable, or Foreman itself can reboot it if it doesn't ask for a 'provision' template in a reasonable amount of time. I realize this can make host creation slower, so maybe we can set this flag (optional) too but not make it mandatory?

Does that sound good @lzap ? My main goal with this is to not complicate the host form further.

ohadlevy: can't we use it per hardware model ?

lzap: > Does that sound good @lzap ? My main goal with this is to not complicate the host form further.

That looks like error prone and fragile workflow to me, I don't like it.
Reasons:
- Users who will have most of their infrastructure on the loader we will
  hardcode to load as the second would need to wait from now on.
- How long should we wait? Some servers takes ages to initialize, this
  will make the proces twice as slow as it was.
- Some servers might get stuck in this reboot loop easily - EFI firmware
  loaders do random things on failures (e.g. halts system completely).
- How about servers without BMC? I assume your suggestion would only
  work with BMC.
- This breaks discovered hosts.

## 

Later,
 Lukas #lzap Zapletal

lzap:
I have updated the RFC according to the current implementation we are merging today. @dLobatog

PR #6, opened 2016-06-28
Added extensible provisioning templates RFC

ShimShtein:


text/0002-extensible-provisioning-templates.md
```diff
@@ -0,0 +1,118 @@
...
+
+As part of the effort to clean up Foreman's core from puppet related code,
+I have stumbled upon `Host::Managed#handle_ca` method. This method is responsible
+for dealing with certificate autosigning mechanism. The method is called from different parts
+in the code: from `UnattendedController` and from two orchestration concerns: `Orchestration::SSHProvision`
+and `Orchestration::Compute`. The motivation for this method is to keep the autosigning window open
+for as shortest period of time as possible.
+My suggestion is to solve this design flaw by using extensible templates in the following process:
+
+1.  Foreman will generate a certificate during the regular orchestration process and store it as part of
```
timogoebel, line 27: Do I understand it correctly, that there will be no more autosigning? How can we make sure the certificate is transferred in a secure way?

ShimShtein, line 27: Yes, no more autosigning.
Two options that I can think of to do it in more or less secure manner:
1. Simple one: render the cert as plain text into the template:

```
cat << EOF > /my.cert
<%= host.stored_certificate %>
EOF
```

It assumes that the template retrieval is secure.
2. embed a secure retrieval mechanism into the template:

```
wget -O /my.cert <%= foreman_url %>/hosts/<%= @host.id %>/stored_certificate?one_time_password=<%= @host.one_time_password %>
```

timogoebel, line 27: > It assumes that the template retrieval is secure.

If I'm not mistaken, it currently is not. That's why I'm asking. :-)

ShimShtein, line 27: Actually I am open to suggestions. I think it's the same problem with any secret that we want to pass to the host.

GregSutcliffe, line 27: It's isn't secure, but technically neither is autosign. Consider today's method:
1. Host GETs /unattended/provision
2. Orchestration of handle_ca is triggered
3. Autosign entry is put into the autosign file

At this point, any attacker watching the network traffic knows the name of the host being built, and can request the cert from the puppet master - it probably has about 15 mins to do so, given a reasonably average Kickstart process

In this approach, we render the certificate in the finish script, which is also requested via /unattended/finish - again, a watching attacker could use this info.

So, we're no _less_ secure this way, but (a) we're can be _more_ secure where https is possible (recent versions of Anaconda _do_ support https), and (b) we make the orchestration of "things that need to happen on other servers" more pluggable.

text/0002-extensible-provisioning-templates.md
```diff
@@ -0,0 +1,118 @@
...
+
+    service network restart
+
+registration:
+
+```ruby
+Pagelets::Manager.add_pagelet(
+  "unattended/host_template",
+  :template_extension,
+  :partial => 'unattended/kickstart/finish/networking',
```
GregSutcliffe, line 101: Presumably this partial (defined above) comes from the plugin?

I was thinking it might be something plugins want to allow the user to contro, or modify - in that case you'd want to refer to an actual Template (via Template.find_by_name or something like that) rather than a static partial. Is that possible in this model?

text/0002-extensible-provisioning-templates.md
```diff
@@ -0,0 +1,118 @@
...
+We will use existing mechanism for extending views called [pagelets](http://projects.theforeman.org/projects/foreman/wiki/Pagelets).
+We will add a call to render pagelets that are already registered each time a template is rendered and pass all the information about the template.
+Any plugin will be able to register its own partial that will be embedded into the final template.
+
+## Changes to the core
+
+1.  Add a call to pagelets to the template sent for rendering in `UnattendedController#safe_render`:
+
+```ruby
+render :inline => "<%= unattended_render(@unsafe_template, @template_name).html_safe %> <%= render_pagelets_for(:template_extension, :subject => @template_object) %>" and return
```
GregSutcliffe, line 53: Minor point, but you probably want a "\n" in there between the templates :)

GregSutcliffe:
Overall, +1. I can see advantages here for plugins like REX (/cc @iNecas / @stbenjam :P) where you want to make sure keys get onto a newly built host without necessarily relying on the user to set up snippets correctly. Discovery could probably find a use for it too, as could Salt?

Very keen to hear other ideas for transporting secrets to hosts - my gut feeling is that the current proposal is no less secure than how we do it today (and effectively identical in the case of REX). Maybe we can do even better though...

text/0002-extensible-provisioning-templates.md
stbenjam, line 27: You can always tell puppet ca to sign the certificate on built url call, that'd reduce the amount of time the window is open. Salt does this now with foreman's callbacks: https://github.com/theforeman/foreman_salt/blob/master/app/models/foreman_salt/concerns/host_managed_extensions.rb#L24-L26

text/0002-extensible-provisioning-templates.md
stbenjam, line 53: It's useful, but we need to tell the plugins what template is being rendered, as we have to handle bash (kickstart), dash (preseed finish), cloud-init (user data), etc.   

Not sure you can really always do this "append" case either. For kickstart, you want want to inject into %post non-chroot, or %post chroot, or even maybe %pre, %packages, etc.

text/0002-extensible-provisioning-templates.md
```diff
@@ -0,0 +1,118 @@
...
+  :partial => 'unattended/kickstart/finish/networking',
+  :priority => 100,
+  :onlyif => ->(subject){ subject.name == 'Kickstart default finish' }) # <- the partial will be rendered only for this template
+```
+
+# Alternatives
+
+[alternatives]: #alternatives
+
+Main alternative will be leaving `#handle_ca` in core and make it extensible. IMHO it will only add to Foreman's technical
```
stbenjam, line 110: Why leave it in? You could move the handle_ca code already to a plugin, without touching anything today. You would still need the extensible templates, but you don't necessarily need to change the fundamental way puppet ca works.  That almost seems out of scope to me...

Orchestration is extensible. I don't do it anymore in Salt as I move to the built callbacks, but here's an example from foreman-salt 1.x: 
https://github.com/theforeman/foreman_salt/blob/SALT-1.1.1/app/models/foreman_salt/concerns/orchestration/salt.rb

It's not very nice, but it's 100% possible to do.  It'd be a lot simpler if we'd move to foreman-tasks/dynflow which has things like [subscriptions](http://dynflow.github.io/documentation/#subscriptions).

text/0002-extensible-provisioning-templates.md
GregSutcliffe, line 53: Actually, you could call a finish template from %post just like we do for preseed - it was suggested on community-templates a while back, but never actually happened.

In any case, the concept of adding a call to render extra commands works just as well in provision templates as it does in finish templates.

text/0002-extensible-provisioning-templates.md
ShimShtein, line 27: It sounds reasonable to me, I wonder why the code does it differently today?

text/0002-extensible-provisioning-templates.md
ShimShtein, line 101: Indeed the partial comes from a plugin. It should be possible to call `#unattended_render` from the partial, so the partial (or helper) can render other template dynamically.

text/0002-extensible-provisioning-templates.md
ShimShtein, line 53: > we need to tell the plugins what template is being rendered

That's the reason you have `:subject => @template_object` - it gives the plugin access to `Template` object, so it can examine it. You can see the example in "Usages" section, I am doing exactly this kind of filtering by using `:onlyif` option provided by the pagelet.

text/0002-extensible-provisioning-templates.md
ShimShtein, line 110: > Why leave it in?

`#handle_ca` should be part of puppet plugin, but it is called from provisioning framework (that should get to a different plugin). I don't want to create dependencies between plugins, I think it should be avoided at all costs.

> You would still need the extensible templates, but you don't necessarily need to change the 
> fundamental way puppet ca works.

It's kind of the other way around - the suggestion for extensible templates comes from the need to get rid of `handle_ca` in core. My main goal is to move puppet out of core. Basically I want to bring foreman core as much closer as I can to a state where the core serves only as a registry and most of the actual work (like provisioning, virtual resources and cfgmgmnt) is done via plugins.

GregSutcliffe, line 110: To add to / clarify that, you need to see this: https://github.com/theforeman/foreman/blob/develop/app/controllers/unattended_controller.rb#L17

That's outside the orchestration system and not pluggable. However, you also can't move it (in it's current form) to the orchestration system - the autosign is a (minor) security hole allowing anyone who knows the host name to request a cert, and should not be left open for potentially a long time (eg user clicks build, forgets to reboot host, autosign entry is present for hours or days).

To generalise the question then - we need a good way to move secrets from other parts of the infra to newly-built hosts, _as close to build-time as possible_. We already do this in the templates for passwords, ssh keys, ect - lets standarise on that and use it for certs too (or come up with something even better).

GregSutcliffe, line 110: What I like about this proposal is splitting up things that are genuinely orchestration (e.g. generating the cert on the puppet master) from the task of transporting the secrets to the host.

Today we hit issues where a template will render fine in preview/spoof mode but fail when actually used because the proxy doesn't have (for example) sudo rights to call puppet cert.

With this proposal, we can move cert generation to the puppet orchestration queue (because that's not autosign, it's still safe) which can fail gracefully just like any DNS or DHCP error - and then we can deal with transporting that cert at build-time.

stbenjam, line 110: > To add to / clarify that, you need to see this: https://github.com/theforeman/foreman/blob/develop/app/controllers/unattended_controller.rb#L17
> 
> That's outside the orchestration system and not pluggable. 

It's totally pluggable.  Add another before filter in a controller concern.

ShimShtein, line 110: It is, but not too developer friendly: if you want to extend `handle_ca` functionality, you have to register the extension in the orchestration (probably twice - for two different types of provisioning)  and then create a new filter in unattended controller.
Doesn't seem too nice...

text/0002-extensible-provisioning-templates.md
stbenjam, line 53: Yea, but the render :inline is still just appending the pagelets at the end, how do you inject the pagelet in a certain part of the template?

text/0002-extensible-provisioning-templates.md
iNecas, line 27: I guess historic reasons?

I'm not sure if this proposal actually simplifies the handle_ca… As you would not need also to solve generating of the keys on the server (or orchestrating the call to CA authority to generate the certs/keys) and storing it somewhere as well on the server. When I have two solutions, where one stores private keys on server and the other one not, guess what my winner is :)

Also, I'm not sure the security is at the same level here: with the autosign case, once the attacker gets his certificate issued before the provisioned host, you will at least get an error on the provisioned host once it tries to get it's certs signed. With the cert-inilne case, the attacker can gain the cert without any error to appear anywhere.

I think the secured storage deserves a RFC on it's own, the generation of certs (if we wanted to go that path) another one.

I think the extendable templates feature makes a ton of sense regardless on how we handle the certs. Would it makes sense to make two RFCs from this and discuss separately the certs part and the templates part?

text/0002-extensible-provisioning-templates.md
ShimShtein, line 53: Three options that I can see, but all of them are complicating the design. I would prefer leaving this design as simple as possible and add this complication only when we see demand to it.

Option 1: Use pagelets inside the template. 
Just add a call to `render_pagelets_for :my_extension_point` at each part of the template you want to extend.

Option 2: Enable extension of snippets too.
I am no quite sure where it should be done, but we can move the appending mechanism closer to the renderer itself, so if the main template calls a snippet - the snippet will be extended the same way.

Option 3: Use filter pattern instead of pagelets.
Basic idea is to let each of the plugins to modify the rendered template before it is returned from the controller. This is a very powerful tool, and personally I would prefer not letting this much freedom to plugins. Plus the development cost of this is bigger.

text/0002-extensible-provisioning-templates.md
ShimShtein, line 27: > When I have two solutions, where one stores private keys on server and the other one not, guess what my winner is

Actually, you don't have to store anything - you can cause the machine to ask the proxy to open the autosign window if you wish - it's all up to you.

> I'm not sure if this proposal actually simplifies the handle_ca

I don't try to simplify the handle_ca, I want to get rid of it.

ShimShtein, line 27: Added a separate [pull request](https://github.com/theforeman/rfcs/pull/7) #7 to track the `#handle_ca` issue.

text/0002-extensible-provisioning-templates.md
ares: I understand the desire to reuse pagelets we already have but...
What if we simply add very similar class like Pagelets::Manager but call it hooks/snippets/whatever Manager. Managers could share some parts (e.g. the priority is useful). During the registration one would simply hook the extension to the identifier. We'd also add the macro so anywhere in the template one could call something like `render_extensions_for(identifier)` and that's it. I guess that's similar to Option 1 suggested here.

Also I think it's better to keep things separated, pagelets for extending html pages, template extensions for extending templates. Even if the implementation can reuse most of the code, the naming is important. Note that not every helper users can use in pagelets would work in templates.

ares:
What's the status of this RFC? I think it combines multiple things - template extending, CA orchestration. It's open for a while, if there's no consensus, should we close it? If there's a consensus, should we merge?

ShimShtein:
CA orchestration has been moved to a different RFC (https://github.com/theforeman/rfcs/issues/7).
This one is about extensible templates.
If you want me to change something in the RFC to reflect this, I will gladly do so.

ares:
Nah, it's more about the discussions, sadly github does not allow hiding obsolete comments. Anyway the status is still unclear to me. From what I can tell, there was no decision on how to address other use-cases than appending (my comment didn't change that status :smile: )

ShimShtein:
@ares, I understand that we already have other use cases that could benefit from this design. Let's put them here, to make the decision easier for others.

ares:
I don't think it's even worth of RFC :-) I simply want to have a tool that allows me to extend the template on any place that template allows. Example where this would be useful - if the provisioning template registered an endpoint in kickstart %post and preseed finish template, plugins could hook their cfgmgmt integration, so e.g. chef and salt would not have to have their code present in core templates. REX could use this to install SSH keys.

PR #7, opened 2016-07-10
Add handle_ca_removal proposal

ShimShtein:
This is a standalone proposal as [discussed](https://github.com/theforeman/rfcs/pull/6#discussion_r68983563) in #6.

### DIFF
--- a/text/0003-remove-handle-ca.md
+++ b/text/0003-remove-handle-ca.md
@@ -1,5 +1,5 @@
 - Feature Name: handle_ca_alternatives
-- Start Date: 2016-07-10)
+- Start Date: 2016-07-10
 - RFC PR:
 - Issue:
 
@@ -33,9 +33,13 @@ First we will modify puppet CA server to support Foreman as an [external policy 
 Next we will add a special controller that will respond to those requests, check for authenticity of the CSR and
 return a result that could be parsed by the script.
 
-\[Optional\]
-We can add information to a node that will be passed as part of the CSR, for example we can use a one-time password that
-will be invalidated upon first CSR validation. This will ensure that each host is registered at most once.
+We will add information to a node that will be passed as part of the CSR.
+The information would be a one-time password set on a Host object and will be invalidated upon first
+CSR validation. This will ensure that each host is registered at most once.
+In order to implement this, we will add the password as part of the PXE template in case of bare metals hosts,
+in the finish template in case of user-data VM's and in the SSH template in case of SSH based provisioning.
+After the password is delivered to the host, we can configure puppet agent to [add it to the CSR](https://docs.puppet.com/puppet/latest/reference/ssl_attributes_extensions.html#aws-attributes-and-extensions-population-example)
+
 
 # Drawbacks
 [drawbacks]: #drawbacks
@@ -45,10 +49,9 @@ This feature is supported in puppet stating from v3.4. If we want to support old
 # Alternatives
 [alternatives]: #alternatives
 
-We can use [extensible template provisioning](https://github.com/theforeman/rfcs/pull/6) if we want puppet plugin to
+1. We can use [extensible template provisioning](https://github.com/theforeman/rfcs/pull/6) if we want puppet plugin to
 generate a certificate upon finish template generation.
+2. We can use the model used in salt to open the autosign window on changes in the [build flag](https://github.com/theforeman/foreman_salt/blob/master/app/models/foreman_salt/concerns/host_managed_extensions.rb#L24-L26).
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
-
-What exactly should be the payload that will enable us to trust the CSR.
--- a/text/0003-remove-handle-ca.md
+++ b/text/0003-remove-handle-ca.md
@@ -1,0 +1,54 @@
+- Feature Name: handle_ca_alternatives
+- Start Date: 2016-07-10)
+- RFC PR:
+- Issue:
+
+# Summary
+[summary]: #summary
+
+Currently 'Host::Managed' contains `#handle_ca` method. As part of puppet separation effort,
+this method should be moved away from the core.
+The problem is that it is called from a different locations in code: once from `UnattendedController`
+and twice from orchestration - from `ComputeOrchestration` and from `SSHOrchestration`. Those parts
+shouldn't be aware to the fact that puppet is present.
+
+I suggest using [policy-based autosigning](https://docs.puppet.com/puppet/latest/reference/ssl_autosign.html#policy-based-autosigning)
+for making Foreman an authority for rejecting irrelevant nodes.
+
+# Motivation
+[motivation]: #motivation
+
+1. It eliminates the need in `#handle_ca` method, so it would not be available to core or other parts of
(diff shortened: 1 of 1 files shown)


text/0003-remove-handle-ca.md
```diff
@@ -0,0 +1,54 @@
...
+First we will modify puppet CA server to support Foreman as an [external policy provider](https://docs.puppet.com/puppet/latest/reference/ssl_autosign.html#enabling-policy-based-autosigning):
+1. Add a simple shell script that will use curl to call foreman server/proxy and pass the CSR to the request.
+2. Mark the script as a policy executable in the puppet.conf
+
+Next we will add a special controller that will respond to those requests, check for authenticity of the CSR and
+return a result that could be parsed by the script.
+
+\[Optional\]
+We can add information to a node that will be passed as part of the CSR, for example we can use a one-time password that
+will be invalidated upon first CSR validation. This will ensure that each host is registered at most once.
```
timogoebel: I think we should definitely use a one-time password here. That adds a lot of security imho. How can we transfer the one-time password to a host in a secure way?

ShimShtein: When dealing with security, I would like to define our threats first. The password is a good solution to _some_ threats, but it's not enough/ineffective for others.
@timogoebel, what threat are we trying to minimize here?

timogoebel: @ShimShtein : I was thinking of preventing leakage of sensitive information stored in a puppet catalog, e.g. passwords.

text/0003-remove-handle-ca.md
```diff
@@ -0,0 +1,54 @@
+- Feature Name: handle_ca_alternatives
+- Start Date: 2016-07-10)
```
GregSutcliffe: incredibly minor typo that needs fixing sometime before merge :P

text/0003-remove-handle-ca.md
GregSutcliffe: @timogoebel that's simply the threat of someone getting a certificate by any means - once they have a cert, they can get a catalog.

I would say the threat model is traffic analysis. There are three paths here:
- A physical (or physical-like VM) will boot, request a template, and continue. This is the weak point - rarely do provisioning systems support HTTPS. Any pre-shared info (or hostnames for traditional autosign) would be visible in the reply.
- A user-data based VM has a similar issue - it requests a finish template. This could be HTTPS, but `unattended_url` isn't that flexible to use HTTP or HTTPS as required.
- An SSH-based VM is probably OK - the VM has a key baked in, and the transfer of templates/data is secure.

So, if the one-time key needs to be present via all 3 methods, we need to figure out how. The first case seems the hardest, but we could potentially embed the key in the PXE template. If we added `foremankey=abcdef` to the APPEND line, it could be queried from `/proc/cmdline` later, right?

This could also be added to the user-data metadata (which is, presumably, going by a secure route to the hypervisor) and to the SSH provision methods. So long as all 3 methods result in the key being in the same place on the host filesystem, further orhestration (puppet, etc) can make use of it.

ShimShtein: This means we need to authenticate the machine that sends the request for autosign. From talking to @ohadlevy and looking into [puppet documentation](https://docs.puppet.com/puppet/latest/reference/ssl_autosign.html#security-implications-of-policy-based-autosigning) we do have a method for authentication of virtual machines, but what about physical hosts?
Also, what do we assume the attacker has? An access to the network?

GregSutcliffe: Scratch the PXE idea - TFTP is hardly secure. Still, the attacker would then _have_ to be on the same subnet as the host, rather than intercepting the kickstart anywhere between the host and Foreman...

GregSutcliffe: The only challenge is physical hosts - one possible idea is a two layer model:
- For most people, the current system is good enough - we just have to not make it worse with whatever we decide to change. Pre-sharing a key via a PXE template is fine in this case.
- For the truly paranoid, the Discovery Image could be used. If a unique salt is embedded into the image (via the rebuild-image method) then both Foreman and the Host can generate the same data from, say, the mac or hostname. This could be used as a key.

It's not ideal, but it might be the simplest without requiring manual intervention at the host.

lzap: We have a token for both PXE and iPXE methods, can't we use that?

lzap:
This is too much puppet for my head - how do you eliminate the `handle_ca` method? From the comments from this method it cleans a CA. Does not make sense to me.

ShimShtein:
`handle_ca` is part of autosigning. If we implement this, we will not use autosigning at all and the `handle_ca` method won't be necessary.

ShimShtein:
Added a bit more details about using a one-time password.

Please note the discussion on [outdated diff](https://github.com/theforeman/rfcs/pull/7#discussion_r70392637) it is not resolved yet.

lzap:
Hmm I have no remarks regarding provisioning, if the CA bits go away from orchestration, that's fine.

GregSutcliffe:
@lzap the reason to ping you was for the thoughts on PXE - if we embed `foreman.secret=somerandomstring` on the PXE APPEND line, then it's accessible in `/proc/cmdline` for use by the installing host. I was wondering how that fitted in with your PXE refactor RFC?

lzap:
> @lzap the reason to ping you was for the thoughts on PXE - if we embed `foreman.secret=somerandomstring` on the PXE APPEND line, then it's accessible in `/proc/cmdline` for use by the installing host. I was wondering how that fitted in with your PXE refactor RFC?

That shouldn't be any problem, but I don't understand why you don't want
to re-use already existing provisioning token which is random enough
(UUID).

## 

Later,
 Lukas #lzap Zapletal

GregSutcliffe:
@lzap good point - more-voices++ :)

@ShimShtein I think the existing token would be sufficient, yes? Do we have any issues with the token timeout? Also, tokens can currently be disabled - if we re-use them for this, I guess we'd have to change that?

ShimShtein:
>  I think the existing token would be sufficient

It should be enough, as long as we invalidate the token (at least for signing) after the first request to sign it. The timeout is actually a good thing - we can shorten the attack window with it.

> tokens can currently be disabled

I am not familiar with the implications of enabled tokens, so I don't have a strong opinion here.

lzap:
> >  I think the existing token would be sufficient
> > It should be enough, as long as we invalidate the token (at least for signing) after the first request to sign it. The timeout is actually a good thing - we can shorten the attack window with it.

They get invalidated the moment host leaves build mode, at the end of
provisioning.

> > tokens can currently be disabled
> > I am not familiar with the implications of enabled tokens, so I don't have a strong opinion here.

We just need to warn users that they are loosing puppet signing when
they disable it. I can't figure out a scenario where you need to disable
tokens anyway.

## 

Later,
 Lukas #lzap Zapletal

fatmcgav:
Has anyone made any progress on this RFC?

As this would massively help with our Terraform+Openstack+Foreman provisioning workflow. See http://projects.theforeman.org/issues/19075 for more info...

PR #8, opened 2016-07-27
Proposal for refactoring id resolvers in hammer

tstrachota:
I'm putting my ideas into RFC to discuss it with wider audience.


dLobatog:
Not the most knowledgeable person about Hammer here, what I read made sense to me though. Seems like it doesn't involve changes in the API or hammer commands themselves. I think it's down to the main hammer maintainers to decide in the end, thanks for keeping us on the loop with the RFC.

akofink:
I like the idea of moving resolvers into separate, override-able classes. I agree this would add greater flexibility when the generated defaults don't quite match.

Currently, you can also specify a method named `create_<plural_resource_name>_search_options`. Would the searchable options also now be a default object which can be overridden?

One concern: in this new framework, would it be possible to differentiate between options and their plural counterparts (i.e. `--organization` vs. `--organizations`)? While I think we should try to avoid confusing options like this, the new framework should allow separate treatment of these options if required.

tstrachota:
@akofink the `create_*_search_options` would be replaced by the resolver objects so it will no longer be needed.

I'm not sure I fully understood your last concern. Handling of plural options remain more or less the same. Parameters `<resource>_id` will still expand to `--<resource>` and `<resource>_ids` to `--<resource_plural>`.
It's might be a good idea to provide separate resolver objects for handling the plural options.

tstrachota:
@mbacovsky may I ask you for review?

mbacovsky:
I like the overall concept of the layer between commands and bindings. It seems it could solve not only the three problems stated in the motivation. It also gives us more room for extending api doc with more meta-data and slimming hammer commands. I can imagine including description of a call output in the new Action object and possibly moving it to the api doc in the future.

tstrachota:
@mbacovsky yes, I think this layer would be a good place to test additional api metadata, that can be moved to the apidoc if it proves to be useful.

tstrachota:
I'm adding one more related item to think of: _Allowed combinations of expanded options aren't obvious_ (http://projects.theforeman.org/issues/16225)

PR #9, opened 2016-08-04
Allow Ansible roles assignment to hostgroup

tstrachota:
RFC for http://projects.theforeman.org/issues/15870

Quite simple one. Sending it via RFC mainly to gather others' ides for future development and opinions on the role ordering part.


text/0005-ansible-hostgroup.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Config groups
+
+The Foreman uses config groups for bundling multiple puppet classes together. We could use them for Ansible roles too.
+
+Alternatives of the mapping:
+
+  1. **1:N** - The same approach as with puppet classes. One config group can have multiple roles.
+  1. **1:1** - A role itself is a bundle of Ansible modules which makes it very similar to config group. Therefore it might make more sense to model it as one config group. One of implementations could be using config groups for modeling roles directly (probably via STI).
+  1. **no config groups** - Don't use config groups for roles at all. Roles have similar purpose as config groups, but there's nothing to edit on them in Foreman (we only import them). Therefore we can keep the both models to live side by side.
```
ares, line 26: I prefer this option. I know there are users that combine puppet and ansible so it might be asked later by some. Config group would allow to classify host in both cfgmgmt tools at the same time. But if someone asks for it, it should be very simple to just add a link between the role and config group.

text/0005-ansible-hostgroup.md
```diff
@@ -0,0 +1,51 @@
...
+
+Existing `AnsibleRole` model will be connected to `Hostgroup` via simple M:N associating model `HostgroupAnsibleRole`.
+
+**Config groups**
+
+We decided not to use config groups at all. The concept of config groups from the Foreman seem very similar to roles in Ansible world. Therefore we took only alternatives 2 and 3 into account. The difference between those two is rather on the implementation level and users shouldn't recognize it. From that perspective it's simpler to use existing `AnsibleRole` and don't add complexity when it's currently not needed.
+
+**Ordering**
+
+In the first iteration we don't plan to enable editing of the roles' order. We can re-visit this issue later with more feedback from users. The default order in a hostgroup will be alphabetical. Sub-hostgroups and hosts will have inherited roles first and then alphabetically ordered additional roles.
```
ares: While I agree with the descision, I don't like alphabetical order too much. This will encourage users to start giving role names like "01-webserver", "02-database". If we later change the way ordering works, number prefixes would have to be confusing. How about just using association creation time or simply id of the record. This way it's still deterministic. As long as we display the resulting order prior running ansible it should be fine IMHO.

text/0005-ansible-hostgroup.md
dLobatog, line 26: 1:N or 1:1 (which is the one that makes conceptually most sense) don't make configuration any easier, so I'd say don't link config groups and roles, it just doesn't make sense and makes things more complicated

dLobatog:
Fine by me, I already spent significant time talking with people in the #ansible IRC channels to figure out ordering. 

General consensus was that it's just not important, and a bad, unsupported practice if you make roles dependent on other roles without explicitly listing these dependencies in the role itself. Please let's not rewrite a scheduler for running Ansible roles, Ansible itself has sorted that out already.

Users should simply define these dependencies in http://docs.ansible.com/ansible/playbooks_roles.html#role-dependencies . We don't have to do _anything_ in foreman_ansible about ordering. However I think it would make sense to show when a role has dependent roles in the UI.

E.g: role 'car' has dependencies wheel, tire, brake (as per the main.yml file), it'd be nice to show in some way that these dependencies will run.

text/0005-ansible-hostgroup.md
dLobatog: Order should be 100% irrelevant, see https://github.com/theforeman/rfcs/pull/9#issuecomment-237551659  - if anyone is doing role dependencies based on ordering, they're doing it plain wrong

ares: while I agree about this being an antipatern, in real world this happens (at least in puppet and chef realms)
so if we always order by id, I'm happy for now and we'll see from user feedback

GregSutcliffe: You cannot order Puppet modules from the ENC - the YAML hash sent to the puppetmaster is unordered (although sorted in the display, this has no effect on the hash itself), and the puppetmaster makes no guarantees about what it does with that hash internally either. That's one of the reasons we haven't bothered adding any ordering to {Host,Config}groups in the past.

GregSutcliffe:
> Order should be 100% irrelevant, see #9 (comment) - if anyone is doing role dependencies based on ordering, they're doing it plain wrong

@dLobatog I agree depending on your classifier to declare ordering is bad - you can't actually do it in Puppet at all (see other comment for details).

> E.g: role 'car' has dependencies wheel, tire, brake (as per the main.yml file), it'd be nice to show in some way that these dependencies will run.

If you figure out how to display those dependencies from the role metadata nicely, lets make it something that works for Puppet too - we've never been able to show that in the UI before. We could parse the module metadata.json file (if present) and display the same kind of data for Puppet modules.

text/0005-ansible-hostgroup.md
```diff
@@ -0,0 +1,51 @@
...
+
+  1. **1:N** - The same approach as with puppet classes. One config group can have multiple roles.
+  1. **1:1** - A role itself is a bundle of Ansible modules which makes it very similar to config group. Therefore it might make more sense to model it as one config group. One of implementations could be using config groups for modeling roles directly (probably via STI).
+  1. **no config groups** - Don't use config groups for roles at all. Roles have similar purpose as config groups, but there's nothing to edit on them in Foreman (we only import them). Therefore we can keep the both models to live side by side.
+
+## Role ordering
+
+Alternatives for ordering the roles:
+  1. Provide just some default predictable order without allowing users to change it.
+  1. Allow ordering in both hostgroups and hosts, order from hostgroup serves as default order in the host form, but it's still editable there.
```
ehelms, line 32: +1

text/0005-ansible-hostgroup.md
ehelms: From my use of Ansible, I absolutely want to order things. Now, this could be from some naive use of them that I am doing when building playbooks for orchestrating which if that is the case hopefully you can show me where I go wrong. My simplish example.

I have a role that configures and sets up Foreman repositories to install Foreman. I have also a role that knows how to install the right packages and run the foreman-installer. I need these to happen as:
- repositories
- foreman_installer

If I can't set an ordering, I can't ensure this. Now, I see this mention of using the role to define dependencies, however, I don't want my foreman_installer role dependent on the repositories role. I may want to run the foreman_installer role on its own because I have slurped in foreman repositories through my Katello and want to configure and install the repositories from there instead of from the standard location.

If I don't have the ability to configure ordering to my needs, how do I achieve mix'n'match role application to my host?

ares: > You cannot order Puppet modules from the ENC

What I mean is that you can always create new module wrapping others which enforces order. That does not mean people do it. Probably because order should not matter but from my experience it makes things easier so people adds explicit order.

dLobatog: @ehelms Write a wrapper role that sets these two as dependencies, it's the way the Ansible manual & mailing list recommend. It's already a solved problem in Ansible itself we should steer clear from.

With a wrapper role you can keep your repositories and foreman_installer roles independent, yet benefit from having them ordered. I would think foreman_installer should be dependent on 'repositories' though, if some kind of repositories are needed, and have repositories 'understand' repos might be set already

ehelms: Again, from my experience, this sounds like I'd end up creating lots of role sprawl just to use this with Foreman. I'd have my normal roles, and then my wrapper roles for various situations instead of just being able to use Foreman to let me order the roles and let the hostgroup serve as the place I define this.

How does ordering on a hostgroup differ from using a playbook and ordering a set of roles? This is how I am looking at this -- ordering equates to specifying a role list on a playbook so I don't have to manage playbooks, I now can manage host groups.

dLobatog: It doesn't differ, but I'm not proposing writing roles in some order in a playbook. What I propose is to use Ansible's feature of specifying role dependencies via the 'dependencies:' ley (http://docs.ansible.com/ansible/playbooks_roles.html#role-dependencies) and let Ansible drive how your roles run. 

Now think the other way around, I'm a user who follows Ansible's recommended practices of defining role dependencies through the feature I linked above,  Foreman should mess up with the order I specified in my dependencies? What if I have:

A - defines dependencies on B via 'dependencies:' in the playbook

Then I choose to run A and B in Foreman - it'd run B -> A -> B? It's a rabbit hole the Ansible team has already dug into - really no reason to do it ourselves. Again I'll come with the ad-hominem - but speak with devs in #ansible see their opinion - I already did and I'm voicing it here, it's Ansible's job to do this ordering.

ehelms: Could we support both through some toggle?

tstrachota: While I'd find it natural to order roles in the hostgroup/host, @dLobatog put very valid argument about the dependencies. I agree with leaving the ordering off.

tstrachota:
I updated the RFC and removed ordering of roles.

dLobatog:
I don't see a lot of discussion here - I'll merge in 1 week unless there's any dissent here in that period of time

dLobatog:
Merging as this was implemented in https://github.com/theforeman/foreman_ansible/pull/30

dLobatog:
Merged and rebased in 1 commit as 58ad1baa7b9f64ce79736d2f1f379b69853b44d2, thanks @tstrachota and everyone else involved

PR #10, opened 2016-08-04
Add Reactjs to foreman technology stack

gailsteiger:


0000-frontend-development-infrastructure-and-workflow-changes.md
```diff
@@ -0,0 +1,73 @@
+- Feature Name: frontend-development-infrastructure-and-workflow-changes
+- Start Date: 2016-07-20
+- RFC PR:
```
ohadlevy: thats: https://github.com/theforeman/rfcs/pull/10

0000-frontend-development-infrastructure-and-workflow-changes.md
```diff
@@ -0,0 +1,73 @@
...
+
+# Summary
+[summary]: #summary
+
+Currently, client side javascript in foreman is implemented with native javascript and the jquery library and managed by the Rails asset pipeline. In addition, various unrelated libraries are used for specific tasks, charting for example. This RFC proposes adding the React javascript library to the technology stack via NPM and Webpack. [(see Webpack PR)](https://github.com/theforeman/foreman/pull/3433) 
+
+# Motivation
+[motivation]: #motivation
+
+The current setup is based on software which was introduced many years ago, solved old problems that are no longer relevant and do not address the challenges that frontend developers face today. Many of the libraries we use are no longer maintained. The infrastructure is sparse and the responsibility for enforcing best practices and coding standards falls on the developer.
```
ohadlevy: I would give an example (e.g. jquery was used to abstract different browser implementations)

0000-frontend-development-infrastructure-and-workflow-changes.md
```diff
@@ -0,0 +1,73 @@
...
+Note: for those familiar with Angular 1, React components are similar to Angular directives.
+
+The [Fixes #13424 - WIP c3 patternfly react implementation PR] (https://github.com/theforeman/foreman/pull/3603) provides a POC for integrating Reactjs with foreman.
+
+Top level components added to an HTML page (mounted) with the `ReactDOM` package `render` method. This is the point at which the Rails view passes the baton to Reactjs.
+
+The paradigm for using React components in Rails views implemented in the POC is as follows:
+
+>The `mount_react_component` helper in the `ReactjsHelper` module is used in the Rails view to which the component is to be added. The helper has 3 paramaters: a string identifier for the specific component to be mounted, a jquery selector that is used to find the DOM element into which to inject the component and, optionally, initial data from the controller in JSON format. See *'app/views/statistics/index.html.erb'* in the POC for an example.
+
```
ohadlevy: you can write it so it includes syntax highlight(use "```erb")

``` erb
<div id="statistics"></div>

<%= mount_react_component('StatisticsChartsList', '#statistics', @data)%>
```

0000-frontend-development-infrastructure-and-workflow-changes.md
```diff
@@ -0,0 +1,73 @@
...
+[drawbacks]: #drawbacks
+
+Remaining with the current infrastructure and code as implemented is not viable. 
+
+The options are to stick with the current infrastructure and rewrite current javascript code base or to integrate another library or framework.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Opinionated frameworks such as Angular 1, Angular 2 and Ember were considered. As the author of this RFC has particular knowledge and experience with that framework, Angular 1 was used as a proxy for the opinionated frameworks. The opinionated frameworks provide more built in features including application state management, but at the price of flexibility.
```
ehelms: What is the flexibility price? can you give an example?

ehelms:
Can you comment on the state of third-party components? i.e. how much thirdparty re-use does this afford given the other frontend frameworks we use like Patternfly?

thomasmckay:
One of the benefits of bastion/angular is that the UI is driven entirely from the api, thus having a single controller to fix/improve for both the UI, CLI, and user usage of API. Will react promote the same pattern?

0000-frontend-development-infrastructure-and-workflow-changes.md
gailsteiger: Fixed. That's cool. Thanks

0000-frontend-development-infrastructure-and-workflow-changes.md
```diff
@@ -0,0 +1,75 @@
...
+
+Applying an MV* approach would facilitate enforcement of best practices and coding standards and make the code produced more of robust, maintainable and extensible. It must be noted however that, any solution must 'work well' with the current foreman Ruby/Rails code base and plugins, as there is no intention to abandon it.
+
+The tools in the MV* space are often classified as being opinionated or unopinionated with respect to the amount of flexibility they allow the developer. This often maps to the breadth of the issues the library or framework attempts to deal with.
+
+Reactjs is a component-based rendering engine. As such, it is both narrow in scope and unopinionated. This will facilitate integration with Rails with minimum side effects on the existing Ruby/Rails code base compared to the opinionated alternatives.
+
+React components demonstrate a high level of encapsulation and support dependency injection. As a result, the components are very testable. The implementation of the recommendations in this RFC will be accompanied by the addition of linting and testing to the client side development process.
```
dLobatog: Important point (testing) considering our current state with regards to JS testing (we just test through expensive, flaky, integration tests)

0000-frontend-development-infrastructure-and-workflow-changes.md
```diff
@@ -0,0 +1,75 @@
...
+
+``` erb
+    <div id="statistics"></div>
+         
+    <%= mount_react_component('StatisticsChartsList', '#statistics', @data)%>
+```
+
+> The helper injects inline javascript code into the Rails view which calls the `mount` method on the `reactMounter` object exposed on the global name space with the the parameters described above.
+
+> The `reactMounter` object manages a collection of outwardly facing components. The `mount` method finds the markup for the component specified by the identifier, adds the initial data (if provided) as props, finds the DOM element specified by the jquery selector and calls the `ReactDOM render` method with the component markup and a reference to the DOM element as parameters.
```
dLobatog: I think a small paragraph about retrieving data by components that are not written by us would be useful

0000-frontend-development-infrastructure-and-workflow-changes.md
```diff
@@ -0,0 +1,75 @@
...
+
+> ReactDOM adds the component to the HTML page and Reactjs takes over from here.
+
+With this paradigm a contributor working in Ruby/Rails could use a previously developed component without great difficulty.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Remaining with the current infrastructure and code as implemented is not viable.
```
dLobatog: Kind of an extreme statement, why not (I do support the rfc though)? `npm` allows us to use external libraries with ease & we could start testing our current JS with Jasmine, Teaspoon or whatever

0000-frontend-development-infrastructure-and-workflow-changes.md
```diff
@@ -0,0 +1,75 @@
...
+Remaining with the current infrastructure and code as implemented is not viable. 
+
+The options are to stick with the current infrastructure and rewrite current javascript code base or to integrate another library or framework.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Opinionated frameworks such as Angular 1, Angular 2 and Ember were considered. As the author of this RFC has particular knowledge and experience with that framework, Angular 1 was used as a proxy for the opinionated frameworks. The opinionated frameworks provide more built in features including application state management, but at the price of flexibility.
+
+Since Reactjs is less intrusive and does not require a full commitment, it was decided that it was the best solution for our needs.
```
dLobatog: Another drawback from JS frameworks is that they usually have a steeper learning curve, and React is fairly easy to understand - which matters from the community contributor point of view.

0000-frontend-development-infrastructure-and-workflow-changes.md
dLobatog: Angular 1 is deprecated too, and 2 is in beta - migration seems complicated from a random internet search. I've also found React significantly easier to debug (anecdotal evidence). 

Another factor to consider is that it's relatively easy to move away from React to other frameworks, but not the other way around. e.g: Angular closes the door to moving to Ember or whatever else, React can just be used for components in any of these (I suppose that'd be the case for Katello)

0000-frontend-development-infrastructure-and-workflow-changes.md
dLobatog: Another motivation - faster JS development pace, with hot reloading, 'time travel' & so forth

0000-frontend-development-infrastructure-and-workflow-changes.md
dLobatog: React is also pretty minimal in size compared to Angular 2 and Ember (5 times smaller) which is not only faster but it'd be easier to understand the internals if we need to

dLobatog: Another drawback of Angular is its HTML-centric design, where our HTML code would change significantly just for the sake of a JS framework, while non React-related HTML code doesn't have to have any shims like 'ngXXX'

dLobatog:
@thomasmckay I think it it wouldn't, React is purely concerned about the views (and not all of them, only the ones it wants to add sprinkles on).

Thanks @gailsteiger , I've put a few comments inline with questions and various suggestions. What @ehelms said above is an important point IMO. I spoke with multiple frontend developers a while ago at MIQ summit and all of them were concerned about using angular-patternfly (a common point is that it's v1 Angular), in fact https://gerrit.ovirt.org/#/admin/projects/ovirt-engine-dashboard uses React components with Patternfly. It seems the only reason the angular-patternfly project started like that is because there's a Patternfly team of devs who knew and hired for Angular, but Patternfly users themselves are not adopting it as much apparently.

0000-frontend-development-infrastructure-and-workflow-changes.md
lzap: Please name them all, just facts here.

0000-frontend-development-infrastructure-and-workflow-changes.md
```diff
@@ -0,0 +1,75 @@
+- Feature Name: frontend-development-infrastructure-and-workflow-changes
+- Start Date: 2016-07-20
+- RFC PR: 
+- Issue: 
+
+# Summary
+[summary]: #summary
+
+Currently, client side javascript in foreman is implemented with native javascript and the jquery library and managed by the Rails asset pipeline. In addition, various unrelated libraries are used for specific tasks, charting for example. This RFC proposes adding the React javascript library to the technology stack via NPM and Webpack. [(see Webpack PR)](https://github.com/theforeman/foreman/pull/3433)
```
lzap: Is NPM/Webpack required in order to implement React?

We usually prefer evolution patches to big bangs. Authors are asked to break huge PRs into smaller pieces anyway.

lzap:
I still don't have complete idea how this works, but will it work with plugins? Is there any added value for plugins when using React?

And just for case, are we talking about this? https://facebook.github.io/react/

0000-frontend-development-infrastructure-and-workflow-changes.md
dLobatog: It's not , but we had an RFC and a long discussion to discuss NPM/Webpack, it was an unanimous decision

0000-frontend-development-infrastructure-and-workflow-changes.md
dLobatog: jnotify, flot-rails, multi-select-rails (the non-rails one isn't unmaintained), gridster js is abandoned too.. there might be more - another issue is that new JS libraries aren't packaged anymore for Rails, as people tend to use npm to get them in any language

0000-frontend-development-infrastructure-and-workflow-changes.md
tstrachota: That's because React gives you only components. It's only the view layer, not full MVC framework. That's why the related PR also brings in Flux. (I'm not saying it's good or bad, just wanted to point out we're comparing a bit different things here.)

@dLobatog btw did you try to use React components in Katello as a part of your POC? Being a fan of React I read couple of blogposts about integrating React and Angular, but I'm curious if somebody actually already tried that in Katello.

ehelms: I have to defend Angular a bit given I've used it for years with great success. Angular 1 isn't deprecated they are still doing releases and supporting it. Angular 2 is on RC5 which is a bit more advanced than a beta phase (given typical usages). Migration is backed by a documented process and tooling as well as side-by-side ability to run 1 and 2.

I think a valid point worth understanding is @tstrachota mention that there are Angular libraries to use React components so there would be opportunities for sharing and re-use.

0000-frontend-development-infrastructure-and-workflow-changes.md
ehelms: A drawback to some is a benefit to others. From my many years of writing UI code, I love that Angular is HTML centric and does not try to abstract that away. The HTML serves as a clean representation of what you are trying to build view wise and then you put it clear points of action. All of the "shims" are still valid HTML and can be re-used outside of Angular if you wanted. With React, you'll be removing that HTML and moving it into the JS code anyway so your HTML will be changing either way. I'm assuming you don't have to use JSX with React so that becomes a bit of a design choice. Either way, things change.

thomasmckay: Speaking of "community contributor", a long standing blocker for having foreman adopt angular was that deface is not usable. Is react compatible? I'd like to see the deface capability, whether it is deface or something else, part of the design and implementation doc. There are a large number of cases where plugins wish to extend and override the UI. Thanks!

0000-frontend-development-infrastructure-and-workflow-changes.md
gailsteiger: @thomasmckay What you are describing is a server implementation issue. The React implementation should have no knowledge of how the api is implemented on the server. It needs an end point and an understanding of what is in the server response. 
Have you seen [Ohad's PR](https://github.com/theforeman/foreman/pull/3691)

0000-frontend-development-infrastructure-and-workflow-changes.md
gailsteiger: Let me clarify what I mean by outwardly facing components (my terminology.) This is a top level component (basically a container) that is added to the HTML with ReactDOM.render. Components not added directly to the page with ReactDOM.render are not outwardly facing. 

The collection of outwardly facing is not necessarily limited to components written by us so data could be passed as props in the same way. That said, I expect that mounting 3rd party components directly into our HTML would be difficult to implement without encapsulation and single responsibility issues. My intuition suggests wrapping the 3rd party component in a component we write that would set up the connections to the app.

Let me also add that the above mentioned paradigm is a bit of an odd duck. foreman is not a single page app and it relies almost exclusively if not exclusively on server side rendering via erb. The purpose of the mounter is to allow us to go forward while not forcing us to make major changes in the structure and functioning of foreman. In the more classic React case, most of the data would be retrieved via Ajax calls.

0000-frontend-development-infrastructure-and-workflow-changes.md
gailsteiger: @ehelms Flexibility in the case that no specific  comprehensive solution fits the requirements of the problem flexibility is the possibility of tailor a solution to the problem which is composed of several separate tools. (Can be risky in itself and would also require more skilled and experienced developers.)

Flexibility also pertains to how well the library/framework works with others and the extent to which the library/framework takes over the application.

No disrespect meant to angular. I've done 3 very successful greenfields projects in angular 1. None of those projects had server side rendering.

Foreman is going to be a server side rendered, not single page, Rails application  for a very long time if not forever. When I reason about using angular in this situation I can't come up with a paradigm that I think would be satisfactory.

Let's say I want to add a dynamic, interactive, client side list to an existing HTML page. With the React mounting paradigm I can compose an outwardly facing component in ES6, use the helper and mount that component. 

Let's try the same thing with angular. If I haven't already done so I will create a module and define a controller in the module which will manage a portion of the HTML page. That controller is associated with that portion of the page with a ng-controller directive. Then I will either have to write some angularized erb or use ng-include to insert a template. If I use a template I'm going to add template caching to the building and for testing. I  don't think that code would be very testable in any case.

gailsteiger:
@lzap Yes, that's the library we are talking about.

0000-frontend-development-infrastructure-and-workflow-changes.md
gailsteiger: @dLobatog faster development experience with TDD and not having to run the app or use a browser.

ohadlevy: @thomasmckay ReactJS does not enforce you using official API controller vs UI specific controllers. I think this is a different discussion and should not be part of this RFC.

having said that, it many cases it make sense to default to / reuse existing API controllers.

0000-frontend-development-infrastructure-and-workflow-changes.md
ohadlevy: I will use different words, but I think the end result is the same, keep on using our current UI development process for new featured based on previous experience (e.g. for network interfaces, puppet class parameters, smart proxies or remote execution) is going to be fairly (or even extremely) painful.

0000-frontend-development-infrastructure-and-workflow-changes.md
ohadlevy: with html based rendering (e.g you initial component starts in erb) you can still use deface to override it.

further, because of how react is component based, you could deface it to use another react component that reuses the existing components yet add your own specific parts.

but i think the main point is that we don't default to override the whole page, meaning you can use the component initially for very small parts, and if it naturally evolve, then it most likely take bigger parts of the UI, where we would need to address a proper extension points.

IMHO, all the common places where plugins extend the UI (hosts show/edit/list etc) should be discussed in term of how to allow proper API, the long term idea with deface was always lets start by allowing deface, and if something is too common, create an API for it (for example, see smart proxy pagelets or dashboard widgets as an API).

0000-frontend-development-infrastructure-and-workflow-changes.md
ohadlevy: I'll also add that its really hard to use 3rd party react components without a proper package manager such as npm.

0000-frontend-development-infrastructure-and-workflow-changes.md
gailsteiger: @tstrachota That is correct. React will need to be supplemented by state management functionality. The PR uses flux but that is not part of this proposal. Still want to check out various options, including redux. It just seemed like it was time to start the conversation about react.

tstrachota: Personally I've only used Facebook's Flux so I can't say much about other implementations but [Alt.js](http://alt.js.org/) seems to be worth checking too.

0000-frontend-development-infrastructure-and-workflow-changes.md
gailsteiger: What I meant is that we cannot continue to develop JavaScript in the way we have been doing, polluting the global namespace, etc.

dLobatog:
As long as we follow this chart we're good here :smile: 

https://camo.githubusercontent.com/cb16ea37cb25a9e98e2e203bf245ed7d5d379676/68747470733a2f2f63646e2e7261776769742e636f6d2f67616561726f6e2f72656163742d6d616b65732d796f752d7361642f313337376236613663646336343461646665613662663233386630366337356433336564366631652f666174696775652e737667

0000-frontend-development-infrastructure-and-workflow-changes.md
tbrisker: I agree the statement is quite extreme, we could continue using the current code at the cost of reduced development speed and developer happiness. It would make sense to rephrase and explain why this is a Bad Idea(tm) rather then just saying it isn't viable.

0000-frontend-development-infrastructure-and-workflow-changes.md
```diff
@@ -0,0 +1,75 @@
+- Feature Name: frontend-development-infrastructure-and-workflow-changes
```
tbrisker: Shouldn't this be renamed to `Add Reactjs to foreman technology stack`?

0000-frontend-development-infrastructure-and-workflow-changes.md
tbrisker: Since Webpack has been merged already, I suggest updating this to reflect it.

gailsteiger:
Relevant to the above discussion I thought you would be interested in knowing that the Patternfly team is giving serious consideration to creating core PatternFly Web Components which could then be used with different JavaScript frameworks. This would make them framework independent and would be great for reuse-ability. For more about this, see PatternFly team member Dana Gutride's [article](http://developers.redhat.com/blog/2016/08/09/are-web-components-in-the-future-for-patternfly/) on the Red Hat Developers Blog.

dLobatog:
Just an anecdote but I asked around yesterday about Patternfly heatmaps in the #patternfly IRC (https://www.patternfly.org/pattern-library/data-visualization/heat-map/) , they said they have no plans on making any part of the framework 'angular-only', but rather the opposite. It's a fact that some of their components (like that one) are angular-only now though!

ohadlevy:
please rebase this pr, it can now be merged as https://github.com/theforeman/foreman/pull/3603 was already merged.

ohadlevy (approved):

ohadlevy:
@gailsteiger can you rebase please? It should contain only your commits.. thanks :)

gailsteiger:
@ohadlevy rebased

gailsteiger:
@ohadlevy @dLobatog @tbrisker 
Could this be merged?

ohadlevy:
@gailsteiger this from some reason shows up as a commit from @GregSutcliffe, try something like:

```
git fetch origin
git rebase -m origin/master
```

then look at git log, if the commit is not yours, try:

```
git commit --amend --author 'Gail Steiger <gsteiger@redhat.com>'
```

gailsteiger:
@ohadlevy now I have 10 commits

ohadlevy:
merged as f9d5213 thanks @gailsteiger

PR #12, opened 2016-08-21
Open host info to plugins

ShimShtein:


GregSutcliffe:
2 questions:

1) How do we declare ordering? If one plugin should modify the host-info before or after another plugin, how can we declare dependencies? Should this be global (like the numerical values in the orchestration) or something a user should control?
2) Presumably we can call other stuff here too? It seems I could move foreman_default_hostgroup to use this, which would be able to check/set the hostgroup before the classification is returned (and means I don't have to use alias_method_chain anymore :P)

Overall, +1 to giving this option to plugins.

ShimShtein:
> how can we declare dependencies?

I hope we don't have dependencies between plugins, at least at data level. The framework will perform deep merge, so if you want to add more info that should be under a node that is created by some other plugin (for example SCAP that wants to add some info under classes), the plugin is responsible to build the hierarchy itself (so SCAP plugin should create class nodes too). Two plugins can contribute the same node - it's not a problem.

I would be glad to see a use case where we need this collection ordered.

> Presumably we can call other stuff here too?

Do you mean a provider that is supplied by the core? [Like this](https://github.com/theforeman/foreman/pull/3701/files#diff-fc693b404f028a834b1b5a4601eca6f1R9)? This PR is an older version of the work that I suggest here. It's basically the [concern alternative](https://github.com/theforeman/rfcs/pull/12/files#diff-d8a74e643a869e3695c99138c72be023R159) from the current RFC.

text/0012-open-host-info-to-plugins.md
```diff
@@ -0,0 +1,182 @@
...
+  def initialize(host)
+    @host = host
+  end
+
+  # Will return provider-specific portion of the info hash.
+  def host_info
+    throw 'Provider is not implemented'
+  end
+
+  protected
```
ares, line 80: why protected? (I don't want to do code review here, just curious if there's some reason I didn't get from the text

ares:
> Two plugins can contribute the same node - it's not a problem.

As long as one of the values is not scalar, but that's fine I think. I can't find the ordering use case either. Maybe if Katello wanted to add some information to openscap policies so the scanner would know where to pull policy from. But it does not work this way, so it's totally artificial.

I like this approach.

GregSutcliffe:
> I hope we don't have dependencies between plugins, at least at data level

Fair enough. In my head, this is something like `alias_method_chain` and there you always need to be careful about playing nice with others who may be aliasing the core method too. I admit, I can't see any obvious places where it'd be necessary today.

> Do you mean a provider that is supplied by the core?

Yes and no. `default_hostgroup` is currently implemented as an `alias_method_chain` hack on `Host.import_facts`, so that when Foreman gets facts from a Host, we can set the Hostgroup _before_ the ENC data is returned. In my head, it's more natural to do this as part of the ENC retrieval itself, and this PR opens up the option of doing so. However, it's modifying the state of the Host, which may not be a good idea in a system designed to purely _read_ ENC data. So, I was gathering info on (a) if what I want is a good idea, and (b) would it be possible?

I think you've answered B and the answer to A might well be out-of-scope for this RFC :P

text/0012-open-host-info-to-plugins.md
ShimShtein: To reduce the public interface to a minimum. The `:host` attribute is there for the implementers to be able to access the host. It's part of the class context, rather than a property that it manages.

ares: I agree, which is why I'd prefer private :-)

ShimShtein: Speaking in C#/Java terms, this is an abstract class - we want the context (the state) to be available to inheritors.

ares: Still don't understand, each inheriting class will have access to it's own host even if it's defined as a private in their parent. I don't see a reason why they should have access to other providers hosts.

ShimShtein:
> Yes and no. default_hostgroup is currently implemented as an alias_method_chain hack on Host.import_facts, so that when Foreman gets facts from a Host, we can set the Hostgroup before the ENC data is returned.

It is a bit out of scope here, but just to give you some ideas: I am not sure if provider classes is a good solution in that case. I would split that case into two - data that is extrapolated from the ENC and is relevant to a single aspect/plugin, and changes to core/abstract objects like the hostgroup. For the first type of changes we should have a solution in form of facets - each facet will get an opportunity to fill its own data from the facts. The second type is a bit problematic since it is prone to conflicts, we will need to think about a conflict resolution mechanism when we tackle this problem.

text/0012-open-host-info-to-plugins.md
ShimShtein: Right. Took the time to read about private/protected model in Ruby, and apparently you are right. It should be private.
Probably I took the concept from other languages, where protected members could not be accessed from siblings, [like here](http://stackoverflow.com/questions/11631777/accessing-a-protected-member-of-a-base-class-in-another-subclass)

ares: :+1:

ShimShtein:
PR https://github.com/theforeman/foreman/issues/3701 has been updated according to this RFC.

PR #13, opened 2016-08-29
Unify Katello and Bastion UI

waldenraines:
Combine [Katello](github.com/katello/katello) and [Bastion](github.com/katello/bastion) into a single UI project.


text/0000-unify-katello-ui.md
```diff
@@ -0,0 +1,78 @@
...
+
+Keep Bastion and move all AngularJS based Katello UI code into Bastion
+
+### Pros
+- UI code and API code is separated
+- Potential to remove rails from the UI code
+
+### Cons
+- Still have two repositories/projects to manage
+- Would have to release the UI code to see changes in the main project
```
jlsherrill, line 48: I think this would cause more instances of double PRs, not less

ohadlevy:
Whats the biggest pain point that you are trying to solve?
is it workflow related (e.g. where code live)? or rather expectation from other developers to write UI code? I could see the benefit to unify bastion and katello UI (as its used by the same developers)

text/0000-unify-katello-ui.md
waldenraines: > I think this would cause more instances of double PRs, not less

Can you elaborate please?

waldenraines: > I think this would cause more instances of double PRs, not less

Agreed, will add that to the cons for this approach.

waldenraines:
> Whats the biggest pain point that you are trying to solve?
> is it workflow related (e.g. where code live)? or rather expectation from other developers to write UI code? I could see the benefit to unify bastion and katello UI (as its used by the same developers)

The main problem is that it's not always clear to those unfamiliar with Katello UI whether something lives in Bastion or Katello and that oftentimes changes are required in both repositories in order to implement a feature.

Also, Katello is the sole consumer of Bastion making it a somewhat unnecessary and meaningless abstraction.

text/0000-unify-katello-ui.md
```diff
@@ -0,0 +1,78 @@
...
+
+Eliminate Bastion entirely and move everything in Bastion back into Katello.
+
+### Pros
+- One less repository to manage
+- Would no longer require the release of a Bastion gem to update UI code in Katello
+- UI changes would be easier because they would only be made in one repository/project
+
+### Cons
+- Decreases separation of concerns, the UI would be bundled with the API code (although you could count this as a pro
```
ehelms, line 35: I'd add that another con is that whenever you want to make a UI code change, the entire test suite has to be run making UI changes much slower to do.

text/0000-unify-katello-ui.md
```diff
@@ -0,0 +1,78 @@
...
+since it would bring us into consistency with [Foreman](https://github.com/theforeman/foreman))
+
+## Option 2:  Move Katello UI code into Bastion
+
+Keep Bastion and move all AngularJS based Katello UI code into Bastion
+
+### Pros
+- UI code and API code is separated
+- Potential to remove rails from the UI code
```
ehelms, line 44: Test runs are quicker and more targeted when changes are made (and in theory then quicker to make)

daviddavis: Could you expand on these pros? I'm not sure I see how having the API and UI in separate repos (and removing rails) is a benefit? I guess I can see how they would provide some [more] separation than we already have now?

waldenraines: The only benefit from my perspective of the separation of the UI and the API is that then we could remove the need for rails in the UI.  This is a huge benefit to UI work because rails weighs a ton and running straight static assets locally while hitting an API on some other sandbox (with the requisite disk space and memory) would be a win from my perspective.  No need to run a suite of rails tests before checking in code, etc.

However, most times that we work on a feature we also need to alter the APIs so this could all be a wash.  Although as our APIs become more stable this may change.

I don't necessarily agree with this pro (or even this approach), I just put it in here for completeness.  I definitely do not agree on separation for separation's sake or to get a ✔️  on some imaginary dev report card.

text/0000-unify-katello-ui.md
waldenraines: Do you mean the entire UI test suite or the rails tests too?

text/0000-unify-katello-ui.md
ohadlevy: cant we conditionally run tests based on the directory/files changed? e.g I could see how the angular tests are running only when bastion is changing and vice versa, I dont see that as a reason to split into multiple repositories.

ekohl: Pytest has [markers](http://docs.pytest.org/en/latest/example/markers.html#mark-examples) which allow you to run a subset of the tests. Maybe that could help?

[edit]: assuming there's a ruby equivalent

text/0000-unify-katello-ui.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Cons
+- Still have two repositories/projects to manage
+- Would have to release the UI code to see changes in the main project
+
+## Option 3:  Move API code into an API project, combine Bastion back into Katello
```
jlsherrill, line 50: To make sure i understand this right, Option 2 and 3 are really the same thing, just with different repo names?  

In both cases all the UI code is in one repo and all the model/api code is in a second repo?

waldenraines: Yeah, I guess that is true for the most part.

The main difference is that in option 3 the code and configuration for displaying web pages to the browser would be in the UI project whereas the API would be only things needed for an API and AR and it would be able to stand alone from the UI.

I think in option 2 the katello project would consist of the API plus some other UI specific code or configuration.  I may be wrong about this and you'll have to pardon my lack of knowledge about engine configuration maybe @ehelms or someone else can chime in here.

Other than that the project names would match up better in option 3 in my opinion and it may be easier pull out the API from katello than it is to pull out the UI, I am not sure.

I can combine these two options into a 2 a,b on the next revision of this RFC if needed.

text/0000-unify-katello-ui.md
waldenraines: > the entire test suite has to be run making UI changes much slower to do.

Six seconds is "much slower"?

```
[walden@[redacted] bastion (master)]$ time grunt test
Running "karma:unit" (karma) task
13 09 2016 11:16:33.236:INFO [karma]: Karma v0.13.22 server started at http://localhost:9876/
13 09 2016 11:16:33.245:INFO [launcher]: Starting browser PhantomJS
13 09 2016 11:16:33.794:INFO [PhantomJS 2.1.1 (Mac OS X 0.0.0)]: Connected on socket /#t9wROIfxerGSY4eaAAAA with id 10996161
PhantomJS 2.1.1 (Mac OS X 0.0.0): Executed 178 of 178 SUCCESS (1.564 secs / 1.728 secs)

Done, without errors.

real    0m5.770s
user    0m5.743s
sys 0m0.554s
```

In other words, we'd only be adding six seconds of tests by moving bastion back into katello.

ehelms: I meant the entire test suite of checks we run on Katello today. Not that PRs would get slower, that one con to having the different concerns together is that pure JS changes require the Ruby tests to run making them slower to iterate.

text/0000-unify-katello-ui.md
ehelms: I think as long as we can smartly detect for a patch where the changes are to determine which tests to run we could.

text/0000-unify-katello-ui.md
ehelms: For me, this one doesn't get us much, as I would think if anything the "base" project is the core code + APIs. With that, you can do anything - run it headless as a pure API server that you automate against or spinning up the UI aspect.

text/0000-unify-katello-ui.md
```diff
@@ -0,0 +1,78 @@
...
+be three potential UI repositories to alter in order to complete new features.
+
+The expected outcome of this feature is to reduce the complexity of working on UI code in Katello by combining projects.
+
+# Detailed design
+[design]: #detailed-design
+
+## Option 1:  Move Bastion code back into Katello
+
+Eliminate Bastion entirely and move everything in Bastion back into Katello.
```
ehelms, line 27: Would you see this as merging what is currently Bastion into `bastion_katello` or plopping it down into the `engines/` directory and keeping the code as is?

ohadlevy:
we've done something similar with Foreman using travisCI, see
https://github.com/theforeman/foreman/blob/develop/script/travis_run_js_tests.sh

text/0000-unify-katello-ui.md
ehelms: This is true, it would become similar to how the CLI is handled today with both relying heavily on the API. Sometimes I think that separation of concern is good for development, and forces developers to think about each aspect and how it would be used. As there is that tendency when they are all together to make one big PR that changes everything, making it harder to review from a length of PR perspective and from a muddying the waters perspective. I have to think about, all at once, what is the API change, how does that work for pure API users. What is the UI change, how does that look and feel, is the UI influencing the API design in a way thats bad for API users.

waldenraines:
Want to bring this back to everyone's attention for another round of comments.

jlsherrill:
+1000 to bringing bastion back into katello, i think at this point it makes the most sense

PR #14, opened 2016-09-12
UEFI HTTP and Grub2 UEFI HTTP booting

lzap:
This RFC does not depend on UEFI work which was merged recently at all, it's
just another way of booting systems without TFTP/PXE.

### DIFF
--- a/text/0000-uefi-http-boot.md
+++ b/text/0000-uefi-http-boot.md
@@ -1,0 +1,75 @@
+- Feature Name: UEFI HTTP Boot
+- Start Date: 2016-09-12
+- RFC PR:
+- Issue:
+
+# Summary
+[summary]: #summary
+
+UEFI supports new kind of boot via HTTP instead of traditional (and slow) TFTP. Implementing this in Foreman is possible with relatively small amount of changes.
+
+# Motivation
+[motivation]: #motivation
+
+Todays firmware is complicated and implementing a HTTP stack is possible, therefore we will see less TFTP in the future. Foreman already supports iPXE HTTP booting which is used for booting virtual machines with iPXE firmware built-in and UEFI HTTP Boot feature will be essentially the same. The feature can enable both UEFI HTTP and iPXE HTTP booting as the implementation requires only two main steps described below.
+
+# Detailed design
+[design]: #detailed-design
+
+In order to support HTTP booting, we need to have an HTTP service deployed. Until now, users were deploying their own services becaue foreman-proxy only provides TFTP service and HTTP API. As part of this RFC, I propose to implement a simple HTTP/HTTPS service as a foreman-proxy plugin serving all files from TFTP directory. Only regular files are supported, listing of directories must be disabled.
+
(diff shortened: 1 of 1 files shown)


text/0000-uefi-http-boot.md
```diff
@@ -0,0 +1,75 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Todays firmware is complicated and implementing a HTTP stack is possible, therefore we will see less TFTP in the future. Foreman already supports iPXE HTTP booting which is used for booting virtual machines with iPXE firmware built-in and UEFI HTTP Boot feature will be essentially the same. The feature can enable both UEFI HTTP and iPXE HTTP booting as the implementation requires only two main steps described below.
+
+# Detailed design
+[design]: #detailed-design
+
+In order to support HTTP booting, we need to have an HTTP service deployed. Until now, users were deploying their own services becaue foreman-proxy only provides TFTP service and HTTP API. As part of this RFC, I propose to implement a simple HTTP/HTTPS service as a foreman-proxy plugin serving all files from TFTP directory. Only regular files are supported, listing of directories must be disabled.
```
timogoebel, line 19: Why not proxy the request to Foreman via "Templates Proxy"-feature but serve the deployed tftp files?

timogoebel, line 19: Can you imagine a real world scenario where https from a self-signed ca would work? Does UEFI firmware check https certificates?

lzap: > Why not proxy the request to Foreman via "Templates Proxy"-feature but serve the deployed tftp files?

We need to server the Grub EFI loader itself anyway, so we could use the same directory for consistency.

lzap: > Can you imagine a real world scenario where https from a self-signed ca would work? Does UEFI firmware check https certificates?

Looks like there is no HTTPS support in the spec, this confirms it: https://firmwaresecurity.com/2015/05/09/new-uefi-http-boot-support-in-uefi-2-5/

But on the other hand iPXE does support HTTP UEFI chainboot, and then you can use https from there on. There is a patch pending from Red Hat engineers into iPXE project to allow embedding your own CA into the iPXE script itself (today it only works you can embed the CA into the firmware itself). This makes it quite usable.

But HTTP -> HTTPS is not super secure way of doing things...

I have no idea how https would work

lzap: FYI http://ipxe.org/appnote/uefihttp

lzap:
We have a report from user lero on IRC that Grub2 can HTTP boot Linux via UNDI/PXE with this statement:

```
linuxefi http://xxxxx/boot/CentOS-7.2-x86_64-vmlinuz
```

The hostname `xxxxx` does not actually mater, it always keeps the same host it TFTP-booted from. Interesting observation. The only thing that blocks users from this are TFTP files exposed via HTTP - and that's exactly what this RFC is all about.

For CentOS6 one needs to use `linux` command instead.

PR #15, opened 2016-09-13
Rename plugin migrations for easier uninstallation

ShimShtein:


GregSutcliffe:
+1

This seems cleaner than the alternative suggested. I think the responsibility on the plugin dev can be mitigated by appropriate updates to the docs and template plugin - I suspect most people just copy what's already there.

dLobatog:
I'm of the same opinion as @GregSutcliffe . It seems very clear that this is a good addition that doesn't need any trade-offs discussion, please open PRs for these changes rather than RFCs. In particular:
- Changes to migrations to the plugins template
- Changes to migrations in plugins
- docs
- Packages can also take advantage of SCOPE=plugin and remove the migrations when uninstalled.

timogoebel:
I agree with the both of you. Let's get started.

text/0015-plugin-migrations.md
```diff
@@ -0,0 +1,61 @@
...
+
+# Detailed design
+[design]: #detailed-design
+
+I want to base my solution on the fact that rails migrations have a notion of
+[scoping](http://edgeguides.rubyonrails.org/engines.html#engine-setup).
+I suggest creating naming convention for migrations added by a plugin: the name
+of each file should look like this:
+
+*timestamp*_*migration_name*.*PLUGIN_NAME*.rb
```
lzap, line 27: Overall I agree, but can you be very explicit about the `PLUGIN_NAME` here? In the past, we had some plugins that were incorrectly named (e.g. discovery vs foreman_discovery inconsistency in constants, plugin name and some namespaces). I think we should state what this tag should match to (is it plugins main module name? camelcase or underscore with downcase?)

I think I can assume `foreman_plugin_name` here, but let's be explicit.

ohadlevy:
+1 to using generators, e.g. in a foreman-devel gem or similar, so we can
auto create these instead on relay on manual renaming.

text/0015-plugin-migrations.md
ShimShtein: From migrations point of view it doesn't matter, but if we want to automate the task, we should probably converge to `foreman_plugin_name`.
Do we have a spec for proper plugin name?

lzap:
> Do we have a spec for proper plugin name?

And you hit the right point :-) Nope.

## 

Later,
 Lukas #lzap Zapletal

daviddavis:
There is some text on the foreman wiki page for creating plugins on how to name your plugin:

http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Naming-your-plugin

ShimShtein:
I have published [foreman_devel](https://github.com/ShimShtein/foreman_devel) plugin that contains a generator for migrations. I kind of avoided the plugin naming convention problem there (at least for now), but thanks for the reminder to look in the ~~FM~~ docs, @daviddavis!
Probably I will use that convention in other generators that I will add to foreman_devel.

PR #16, opened 2016-09-17
RFC: Katello as a Real Plugin

ehelms:
Initial write-up of previously discussed and gathered information regarding Katello as a plugin that can be added to Foreman. Given the breadth and number of discussions on this topic, I have likely missed or overlooked something. However this should start as a good discussion point for those new to the topic and those familiar with it.

### DIFF
--- a/text/0016-katello-as-a-plugin.md
+++ b/text/0016-katello-as-a-plugin.md
@@ -1,0 +1,141 @@
+- Feature Name: Katello as a (Real) Plugin
+- Start Date: 2016-09-09
+- RFC PR: (leave this empty)
+- Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+The goal of this effort is to allow Katello to be treated from a community and technical perspective as any other plugin.
+
+# Motivation
+[motivation]: #motivation
+
+Katello was originally a stand alone Rails application that attempted to integrate with Foreman through a single-sign-on process, Rails engines that knew how to communicate with one another and switching menu structures. After a desire to centralize on a single community and bring integration tighter, Katello was moved to be a plugin but given the constraints on Katello for depoyment along with timeline the deployment of Foreman and Katello was made unique when compared to any other plugin. 
+
+# Detailed design
+[design]: #detailed-design
+
+Given this is a large area of work with many facets, the design will list out each issue with proposed design solutions and any drawbacks. This may be able to be split out to their own RFCs over time.
+
+## Certificates
+
+Foreman defaults to using Puppet certificates for the web UI. Katello requires a certificate authority to hand to Candlepin for cutting client certificates and entitlement certificates. The same certificate authority used to cut client certificates is needed to have signed the server certificates so that clients can talk back to the server and access HTTPs based repositories. Further, certificates are needed to configure communication with Pulp, Qpid, Capsules (Smart Proxies) and creating entities like trustores for Candlepin.
+
+Another aspect of certificates is the desire by users to be able to supply their own certificates for the web UI. Users should be able to change those fluidly without having to propagate new certificates across their entire client infrastructure.
+
+#### Design
+
+Introduce the concept of an infrastructure port that the application runs on and uses the self-signed certificates for. This port would exist in a known range (e.g. 9070 - 9079) so that all other backend services could be similarly configured within this range with well defined port values. For example, Pulp could be deployed on 9072 and Candlepin 9073. All Smart Proxies, Capsules and Clients would talk to port 9070 and leave the web UI to run on 443 as is today. This would allow users to switch out the web UI certificates simply and easily while requiring no updates to the clients. 
+
+Action Items:
+ 1. Complete [PR](https://github.com/theforeman/puppet-foreman/pull/374)
+ 2. Add upgrade flag or steps to upgrade existing Smart Proxy to new port
+
+#### Drawbacks
+
+This could introduce potential performance issues with running the application on two ports or complex configuration to ensure proper redirects. This change would introduce a one-time infrastructure update for existing users to move servers and Smart Proxies to the new port and certificate setup. Clients would also need updating to ensure they are pointing and talking to the correct port with RHSM. 
+
+## Smart Proxy vs. Capsule
+
+A Capsule is essentially a Smart Proxy that deploys Pulp, and a reverse proxy to achieve content federation and an isolated communication point fo clients. Foreman deploys the Smart Proxy on port 8443 by default, while Katello deploys it on 9090 by default. Katello deployed Smart Proxies on 9090 because Candlepin was already running on 8443 on the server.
+
+#### Design
+
+Given the above discussion of infrastructure port, introduce a designated Smart Proxy port (e.g. 9071) that is predictable and not commonly used as 8443 is today. Further, merge the puppet-capsule and puppet-foreman_proxy modules to unify Smart Proxies and Capsules and give the content aspects of a Smart Proxy a new name (e.g. content node or content proxy). Content aspects are the deployment of Pulp for content federation and a reverse-proxy to create an isolation point for clients to communicate through and only through a Smart Proxy.
+
+Action Items:
+ 1. Change nomenclature from Capsule to Content Smart Proxy
+ 2. Merge puppet-capsule into puppet-foreman_proxy
+ 3. Rename Capsule scenario
+ 4. Add installer migration to move old capsule module to smart proxy
+
+#### Drawbacks
+
+The port change would require infrastructure propogation for existing users. The current Capsule concept does not use a Smart Proxy in the traditional sense given it uses a reverse proxy for clients to talk back to the server in an isolated way and deploys Pulp for content federation. Given Pulp has it's own stable API, there is no need to go through the Smart Proxy. However, the Smart Proxy is used to provide statusof Pulp, storage information and indicators of enabled features.
+
+## Puppet Modules
+
+Currently Katello maintains a set of puppet modules that set it a part from other plugins. In order to install and configure Katello, it requires `puppet-katello` for configuring Katello as a plugin and backend services and `puppet-capsule` to configure content features on a Smart Proxy. Further, Katello maintains a set of puppet modules for each of it's backend services including the aforementioned certificate setup.
+
+#### Design
+
+This has been previously discussed and outlined by an existing [foreman-dev post](https://groups.google.com/forum/#!topic/foreman-dev/hebniL0yyqM). The larger proposals would be:
+
+ * Add Katello plugin manifest to puppet-foreman
+ * Add Katello plugin manifests for each backend service to puppet-foreman
+ * Add puppet-capsule features to puppet-foreman_proxy
+
+#### Drawbacks
+
+Adds a lot more manifests to puppet-foreman and puppet-foreman_proxy and keeps things less separated and focused.
+
+## Installer Repositories
+
+The `katello-installer` uses the foreman-installer and scenarios in order to install providing it's own scenario for Katello and for the Capsule. These including the system checks, and hooks are all contained within the `katello-installer` repository.
+
+### Design
+
+Unify the installer by storing multiple scenarios and dependencies in the single `foreman-installer` repository. This would put everything in one place and encourage more design work to support a variety of scenarios for different deployment types.
+
+Action Items:
+  1. Move katello-installer checks to foreman-installer
+  2. Move katello-installer hooks to foreman-installer
+  3. Move katello scenario to foreman-installer
+  4. Move Capsule (or merged scenario from above) to foreman-installer
+  5. Move devel scenario to foreman-installer
+  6. Update Puppetfile for Katello modules
+  7. Move capsule-certs-generate to foreman-installer
+
+### Drawbacks
+
+Merging the two together could muddy the waters with respect to different sets of hooks, requirements and modules required to install a given scenario. 
+
+## Release Repositories
+
+Currently the Foreman repositories are housed at [http://yum,theforeman.org](http://yum.theforeman.org) and Katello's at [https://fedorapeople.org/groups/katello/releases/yum/](https://fedorapeople.org/groups/katello/releases/yum/).
+
+#### Design
+
+Move the Katello repositories to be served from the Foreman repository location. Further, would be to move all packages in katello-packaging to foreman-packaging and into the plugins repository.
+
+#### Drawbacks
+
+None.
+
+## Github Organization
+
+Currently Foreman and Katello are split into two separate github organizations
+
+### Design
+
+Move all active Katello projects into the Foreman github organization and update all related infrastructure:
+
+ * Redmine repositories that are synced
+ * prprocessor repository list
+ * CI jobs
+
+## Documentation
+
+Today the Foreman website and plugins are kept in a different repository than Katello. Further, the websites and documentation are at different URLs [http://www.theforeman.org](http://theforeman.org) and [http://www.katello.org](http://www.katello.org).
+
+#### Design
+
+Action Items:
+ * Move Katello's documentation inside the plugin section of the Foreman's website documentation
+ * Get rid of the katello.org repository and website.
+ * Add documentation to Katello around deployment strategies, diagrams
+
+#### Drawbacks
+
+The effort to re-do the website design, and include all of the current information displayed such as troubleshooting, user guides, and developer information.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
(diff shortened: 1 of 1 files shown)


text/0000-katello-as-a-plugin.md
```diff
@@ -0,0 +1,107 @@
...
+
+Another aspect of certificates is the desire by users to be able to supply their own certificates for the web UI. Users should be able to change those fluidly without having to propagate new certificates across their entire client infrastructure.
+
+#### Design
+
+Introduce the concept of an infrastructure port that the application runs on and uses the self-signed certificates for. This port would exist in a known range (e.g. 9070 - 9079) so that all other backend services could be similarly configured within this range with well defined port values. For example, Pulp could be deployed on 9072 and Candlepin 9073. All Smart Proxies, Capsules and Clients would talk to port 9070 and leave the web UI to run on 443 as is today. This would allow users to switch out the web UI certificates simply and easily while requiring no updates to the clients. Some discussion of this can be seen [here](https://github.com/theforeman/puppet-foreman/pull/374).
+
```
ohadlevy: :+1: sounds like a great idea ;)

text/0000-katello-as-a-plugin.md
```diff
@@ -0,0 +1,107 @@
...
+#### Drawbacks
+
+This could introduce potential performance issues with running the application on two ports or complex configuration to ensure proper redirects. This change would introduce a one-time infrastructure update for existing users to move servers and Smart Proxies to the new port and certificate setup. Clients would also need updating to ensure they are pointing and talking to the correct port with RHSM. 
+
```
ohadlevy: I'm guessing this could be mitigated by not enabling this by default? Also, clients would still communicate to the infrastructure ports, so in general its only the web ui which has two ports by default?

text/0000-katello-as-a-plugin.md
```diff
@@ -0,0 +1,107 @@
...
+## Smart Proxy vs. Capsule
+
+A Capsule is essentially a Smart Proxy that deploys Pulp, and a reverse proxy to achieve content federation and an isolated communication point fo clients. Foreman deploys the Smart Proxy on port 8443 by default, while Katello deploys it on 9090 by default. Katello deployed Smart Proxies on 9090 because Candlepin was already running on 8443 on the server.
+
+#### Design
+
+Given the above discussion of infrastructure port, introduce a designated Smart Proxy port (e.g. 9071) that is predictable and not commonly used as 8443 is today. Further, merge the puppet-capsule and puppet-foreman_proxy modules to unify Smart Proxies and Capsules and give the content aspects of a Smart Proxy a new name (e.g. content node or content proxy).
```
ohadlevy: why not change cadnlepin port? I assume the reason is that you would have to reconfigure all of its clients ? (every machine that uses yum / subscription manager)

text/0000-katello-as-a-plugin.md
```diff
@@ -0,0 +1,107 @@
...
+
+#### Drawbacks
+
+The port change would require infrastructure propogation for existing users. The current Capsule concept does not use a Smart Proxy in the traditional sense given it uses a reverse proxy for clients to talk back to the server in an isolated way and deploys Pulp for content federation. Given Pulp has it's own stable API, there is no need to go through the Smart Proxy. However, the Smart Proxy is used to provide statusof Pulp, storage information and indicators of enabled features.
+
```
ohadlevy: I think we need to think of the smart proxy as two different parts:
1. the smart proxy service it self  (with its api's)
2. the way foreman looks at smart proxies (location, organizations, features etc).

It make sense that every backend service is a smart proxy (from option 2), and probably not every backend needs to be transformed into a smart proxy plugin (1) - yet, in some cases it make sense, it could also be used as a way to act on the backend without a service (think pulp migrations).

ohadlevy: can we also git rid of the notion of capsule in katello? I find it confusing.

text/0000-katello-as-a-plugin.md
```diff
@@ -0,0 +1,107 @@
...
+The `katello-installer` uses the foreman-installer and scenarios in order to install providing it's own scenario for Katello and for the Capsule. These including the system checks, and hooks are all contained within the `katello-installer` repository.
+
+### Design
+
+Unify the installer by storing multiple scenarios and dependencies in the single `foreman-installer` repository. This would put everything in one place and encourage more design work to support a variety of scenarios for different deployment types.
+
+### Drawbacks
+
+Merging the two together could muddy the waters with respect to different sets of hooks, requirements and modules required to install a given scenario.
```
ohadlevy: I would also add, that katello uses the installer in a way most other plugins do not, for example the usage of hooks can actually drive data manipulations on backend services. I would like to use this opportunity to discuss if this is the right way to handle upgrade data transformation (there is already related issue at http://projects.theforeman.org/issues/15866)

text/0000-katello-as-a-plugin.md
```diff
@@ -0,0 +1,107 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+The grand drawback with this work would be that it would interrupt business as usual and no changes or work is needed that would disrupt the release flow of Katello.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+ * What to do about nested Organizations given Katello does not know how to support this given it's organization model.
```
ohadlevy: I would like to ask about a few more items:
1. Allow Katello to operate without candlepin
2. Break the org model between Katello and Candlepin
3. Allow Pulp / Candlpe to run purely on the smart proxy 
4. Review foreman_content (https://github.com/theforeman/foreman_content) for its data model association with foreman objects and content tagging principles, most likely in another RFC...

ohadlevy:
Another point, what about the katello github org vs joining the foreman github org?

text/0000-katello-as-a-plugin.md
timogoebel: How about rewording Capsules to Smart Proxies?

Edit: Should have read this more carefully. There is already a section on this. Sorry.

text/0000-katello-as-a-plugin.md
timogoebel: What do you mean by "content aspects of a Smart Proxy"?

Edit: It gets clear if you read the rest of this RFC. Readers unfamiliar with katello might stumble upon this.

timogoebel:
I really like the ideas in this RFC. I think, it would greatly improve the user experience - especially for new users and really streamlines the setup.

I think, separating candlepin from the repo part of Katello is also a very important topic. Candlepin adds a lot of complexity, that some users might find unnecessary.

One more thing to think about: How can foreman_docker (or any other plugin) use Katello to store docker images in pulp or e.g. show a list of available images from pulp when starting a container?

text/0000-katello-as-a-plugin.md
```diff
@@ -0,0 +1,107 @@
...
+
+None.
+
+## Documentation
+
+Today the Foreman website and plugins are kept in a different repository than Katello. Further, the websites and documentation are at different URLs [http://www.theforeman.org](http://theforeman.org) and [http://www.katello.org](http://www.katello.org).
+
+#### Design
+
+Move Katello's documentation inside the plugin section of the Foreman's website documentation and get rid of the katello.org repository and website.
```
timogoebel: I think, we should add documentation for the different scenarios and some schema drawings on how to set things up. Users might want to set up a ha setup or have special compliance/performance requirements to separate the services. This should be possible and there should be proper documentation.

text/0000-katello-as-a-plugin.md
jlsherrill: clients don't communicate with candlepin directly, so any port change should be fine.  The reverse proxy on the external capsules/smart-proxies does use 8443 iirc though?

text/0000-katello-as-a-plugin.md
jlsherrill: I think all of these items should be separate RFE's.  They really have a different focus than this RFC and are complex in their own right.  I'm fine with taking on item number 1 in this list and opening up an RFC for that.

text/0000-katello-as-a-plugin.md
ehelms: Candlepin port can be changed fairly easily, the reverse proxy for isolation would be an issue as @jlsherrill points out. I also think that in general 8443 is an over used port.

text/0000-katello-as-a-plugin.md
ehelms: That is mentioned here (but could be its own item) -- https://github.com/theforeman/rfcs/pull/16/files#diff-d26c8309f64125dcf90135dc23ccedb5R41

text/0000-katello-as-a-plugin.md
ehelms: That issue seems related to the database more so than the installer. Katello does use the installer in more intricate ways than most plugins, and those concepts would need to be ported to broader concepts that are just in the foreman-installer (e.g. --upgrade, --reset, --upgrade-puppet). I can include more details around that requirement.

text/0000-katello-as-a-plugin.md
ehelms: I tend to agree and was planning to open some additional RFCs as these items get more to functionality and/or changing the architecture of Katello which I think is made easier by first making Katello a regular feeling plugin.

ehelms:
Good point on the Github organization conversation, I'll add that.

ehelms:
```
One more thing to think about: How can foreman_docker (or any other plugin) use Katello to store docker images in pulp or e.g. show a list of available images from pulp when starting a container?
```

Today we have an inverse of this, where Katello supports content types and adds integration into plugins to provide that self-hosted content aspect when Katello is present. Do I read correct you are suggesting the opposite so that a plugin like foreman_docker could instead require Katello to supply Docker content support?

timogoebel:
> > One more thing to think about: How can foreman_docker (or any other plugin) use Katello to store docker images in pulp or e.g. show a list of available images from pulp when starting a container?
> > Today we have an inverse of this, where Katello supports content types and adds integration into plugins to provide that self-hosted content aspect when Katello is present. Do I read correct you are suggesting the opposite so that a plugin like foreman_docker could instead require Katello to supply Docker content support?

Yes. This seems to be the more natural workflow for me. Let's say, as a user I decide at some point: "I want to play with docker.". Then I want the possibility to add "docker support" (the feature) for Foreman. Imho we should think of plugins like new features. The opposite is also important: If I don't want docker support, I want to be able to switch it off.

text/0000-katello-as-a-plugin.md
```diff
@@ -0,0 +1,107 @@
...
+# Detailed design
+[design]: #detailed-design
+
+Given this is a large area of work with many facets, the design will list out each issue with proposed design solutions and any drawbacks. This may be able to be split out to their own RFCs over time.
+
+## Certificates
+
+Foreman defaults to using Puppet certificates for the web UI. Katello requires a certificate authority to hand to Candlepin for cutting client certificates and entitlement certificates. The same certificate authority used to cut client certificates is needed to have signed the server certificates so that clients can talk back to the server and access HTTPs based repositories. Further, certificates are needed to configure communication with Pulp, Qpid, Capsules (Smart Proxies) and creating entities like trustores for Candlepin.
+
+Another aspect of certificates is the desire by users to be able to supply their own certificates for the web UI. Users should be able to change those fluidly without having to propagate new certificates across their entire client infrastructure.
```
dLobatog: This does not mention how to address the 'CA issue'. Couldn't Katello provide the Puppet CA to Candlepin and use that? What changes would it require in terms of puppet modules (i'm thinking on how can we organize all the work there's to do here)

text/0000-katello-as-a-plugin.md
dLobatog: What changes are needed in which modules to do this? I do like the idea very much too :smile:

text/0000-katello-as-a-plugin.md
```diff
@@ -0,0 +1,107 @@
...
+
+#### Drawbacks
+
+This could introduce potential performance issues with running the application on two ports or complex configuration to ensure proper redirects. This change would introduce a one-time infrastructure update for existing users to move servers and Smart Proxies to the new port and certificate setup. Clients would also need updating to ensure they are pointing and talking to the correct port with RHSM. 
+
+## Smart Proxy vs. Capsule
+
+A Capsule is essentially a Smart Proxy that deploys Pulp, and a reverse proxy to achieve content federation and an isolated communication point fo clients. Foreman deploys the Smart Proxy on port 8443 by default, while Katello deploys it on 9090 by default. Katello deployed Smart Proxies on 9090 because Candlepin was already running on 8443 on the server.
```
dLobatog: I suppose we could change the proxy to run on 9090 by default...? Also this point could address that we should standardize on one or the other (I'd prefer smartproxy) across docs

text/0000-katello-as-a-plugin.md
```diff
@@ -0,0 +1,107 @@
...
+
+Currently Katello maintains a set of puppet modules that set it a part from other plugins. In order to install and configure Katello, it requires `puppet-katello` for configuring Katello as a plugin and backend services and `puppet-capsule` to configure content features on a Smart Proxy. Further, Katello maintains a set of puppet modules for each of it's backend services including the aforementioned certificate setup.
+
+#### Design
+
+This has been previously discussed and outlined by an existing [foreman-dev post](https://groups.google.com/forum/#!topic/foreman-dev/hebniL0yyqM). The larger proposals would be:
+
+ * Add Katello plugin manifest to puppet-foreman
+ * Add Katello plugin manifests for each backend service to puppet-foreman
+ * Add puppet-capsule features to puppet-foreman_proxy
```
dLobatog: Where would SELinux/firewall changes for backend services happen? This is missing a section on certs too I think

text/0000-katello-as-a-plugin.md
ehelms: We could try that, but I honestly do not want to tie Candlepin to puppet certs. Whatever we can do to not add any more bonds to Puppet would be ideal IMO.

text/0000-katello-as-a-plugin.md
ehelms: Not sure I follow entirely, are you asking about merging foreman-selinux with katello-selinux?

ehelms:
I pushed a number of updates based on comments and feedback. In some areas, I attempted to add more concrete "Action Item" lists.

text/0000-katello-as-a-plugin.md
ares: :+1:  that might help with updates too, user endpoint would be in maintenance mode, while infrastructure port would be available for services like Pulp

text/0000-katello-as-a-plugin.md
ares: I could see the upgrade that katello does in hooks changed in one of two ways, 1) move all execute calls to puppet module where the rest is 2) extract to separate smart upgrade script that would be triggered from the hook

One advantage of 1) is that it could be used for upgrading katello using puppetmaster infrastructure, OTOH I think it does not help with recoveries from failed updates etc.

2) brings a way to have a better control of update flow, e.g. retry if something failed. It could be run independently but also from the installer run.

text/0000-katello-as-a-plugin.md
ares: Or we can install multiple trusted CA certs. User could provide extra CA cert that would be installed with both Puppet and Candlepin CA certs. This would allow us to easily start using new CA that would be used if users do not provide their own. This would be useful today in scenario where I do not use Puppet nor Candlepin but I still want to have x509 certs.

text/0000-katello-as-a-plugin.md
dLobatog: Katello can block creation of nested organizations and fail on installation if some are found (although I'd rather entirely drop the "feature")

text/0000-katello-as-a-plugin.md
```diff
@@ -0,0 +1,141 @@
...
+
+None.
+
+## Github Organization
+
+Currently Foreman and Katello are split into two separate github organizations
+
+### Design
+
+Move all active Katello projects into the Foreman github organization and update all related infrastructure:
```
daviddavis: Would we also rename repositories (specifically Katello/katello)? Seems like katello should be renamed to `foreman_katello` or something if we really want to be in line with other plugins.

ehelms:
I've updated this to add a directory where each of these items will get broken out into their own RFCs due to the size of them.

ehelms:
After some discussions, I've taken each topic here, as a well as a few others that have come out of these and other discussions and moved them into individual RFCs due to their scope. The central RFC will reference each. For a list of them its easiest to look at the list of PRs:

https://github.com/theforeman/rfcs/pulls

PR #18, opened 2016-10-03
Manageable HTTP Proxies

jlsherrill:


text/0000-manageable-http-proxies.md
```diff
@@ -0,0 +1,65 @@
...
+
+2) For each organization, a default proxy can be selected for use when creating repositories.  The goal of this is to mimic the current behavior where a user really does not have to think about setting the proxy.  Modifying this only affects new products.
+
+3) A default proxy can be set for a product, inheriting from the organization default if one is set.
+
+4) At repository creation time, the user can choose to 'inherit' from the product, or override with a particular proxy or none.
+
+5) Upgrades: If a user has already set a default proxy, we would go ahead and create a proxy, associate it with all of their products and repositories and set their org defaults to this proxy.
+
+6) Capsules: You could associate an HTTP proxy to a capsule and when syncing the capsule would use that proxy for its syncs.
```
timogoebel, line 47: If I read "Capsules" as "Smart Proxies", I do see a challenge here: Smart Proxy settings are not managed by Foreman UI right now. Settings are stored in text files on disk. I am wondering, how should the setting be changed on the proxy?

jlsherrill: Yeah, i used the term capsule mainly because this was concerning repo syncing from the foreman server to the smart proxy (in katello's use case), but you do bring up a good point that other parts of the smart proxy may also need to use a proxy to reach to the foreman server.  

I added it to the RFC as it seemed to make sense initially, but as there tends to only be one proxy for a smart proxy (you aren't likely going to want to use one proxy for this plugin and another for 2nd plugin, nor one proxy to sync one repo, and another for a different repo) maybe it makes more sense as a purely smart-proxy side option that is set via the installer.

text/0000-manageable-http-proxies.md
```diff
@@ -0,0 +1,65 @@
...
+
+# Summary
+[summary]:
+
+Adds HTTP Proxies as a first class citizen that is manageable by users for use with syncing and potentially other things.
+
+# Motivation
+[motivation]:
+
+Currently when synchronizing repositories, users can either sync straight to the content source, or globally use an http proxy defined at installation time.  In many cases, you may want to use one proxy for anything syncing outside your network, but no proxy for internal sources.  There may be other cases where you want to use one proxy for syncing one repo and another proxy for a second repo.
```
ehelms, line 14: Nitpick: s/http/HTTP

ehelms: Are we assuming HTTP in any case mentioned means HTTP and HTTPS proxies? Does auth need to be considered for the proxies?

text/0000-manageable-http-proxies.md
```diff
@@ -0,0 +1,65 @@
...
+
+Allowing the user to define their proxies and select them on a per-repo basis would be extremely helpful in these situations.
+
+In addition it could allow easy use of proxies with capsule syncing (which is a requested feature).
+
+# Detailed design
+[design]:
+
+Most of this could be implemented in a new plugin so that other plugins could use this concept.  Alternatively it could be implemented directly in Katello.  Curious to get thoughts around this, but I could see this being useful for compute resources and the docker plugin.
```
ehelms, line 23: +1, seems generic enough to be it's own entity

text/0000-manageable-http-proxies.md
```diff
@@ -0,0 +1,65 @@
...
+ 
+1) A new navigation item would be added under the 'Infrastructure' tab for one of the following:
+
+ * HTTP Proxies (my current preference)
```
ehelms, line 27: :+1:

text/0000-manageable-http-proxies.md
```diff
@@ -0,0 +1,65 @@
...
+4) At repository creation time, the user can choose to 'inherit' from the product, or override with a particular proxy or none.
+
+5) Upgrades: If a user has already set a default proxy, we would go ahead and create a proxy, associate it with all of their products and repositories and set their org defaults to this proxy.
+
+6) Capsules: You could associate an HTTP proxy to a capsule and when syncing the capsule would use that proxy for its syncs. 
+
+# Drawbacks
+[drawbacks]:
+
+1) Adds a new menu entry (can not think of another place to put it.)
```
ehelms, line 52: I think Infrastructure is a good place to put it.

text/0000-manageable-http-proxies.md
```diff
@@ -0,0 +1,65 @@
...
+
+
+# Alternatives
+[alternatives]: #alternatives
+
+1) Having a checkbox that lets the user not use a global proxy.  Much simpler than the above but much less flexible
+
+2) have a freeform URL, user, and password field on each repository/product (requires user to repeat the same info over and over when creating a product/repo).
```
ehelms, line 60: :-1:

text/0000-manageable-http-proxies.md
```diff
@@ -0,0 +1,65 @@
...
+# Drawbacks
+[drawbacks]:
+
+1) Adds a new menu entry (can not think of another place to put it.)
+
+
+# Alternatives
+[alternatives]: #alternatives
+
+1) Having a checkbox that lets the user not use a global proxy.  Much simpler than the above but much less flexible
```
ehelms, line 58: :-1:

ehelms:
Two things I can think of that might be worth talking about:

 1) Permissions/filters for the proxy and any default roles it might be associated to
 2) How content views and their respective content would be handled

ehelms (changes requested):

text/0000-manageable-http-proxies.md
jlsherrill: Yes, the url for the proxy can be http:// or https:// just like today.  Auth is included, see the below where i talk about the attribute for the proxy (url, username, password)

jlsherrill:
>  2) How content views and their respective content would be handled

I'm not sure what you mean by this?

text/0000-manageable-http-proxies.md
jlsherrill: Yes, should i have renamed this section to 'crappy alternatives' :)

mhrivnak:
Pulp could benefit from a similar feature, and it's something I've thought of from time to time. We are planning to at least unify download settings in pulp 3, so they aren't plugin-specific.

Any thoughts on how it would be useful to support this proxy feature in pulp?

ehelms:
Ignore this, long day and I'm not sure what I meant either.

text/0000-manageable-http-proxies.md
ehelms: Touche, I did miss that down there.

text/0000-manageable-http-proxies.md
timogoebel: +1, compute resources would be a great addition. Imagine a corporate environment where your Foreman/Katello server is in an internal network and you want to talk to - let's say - the AWS API which is on the dangerous and evil internet. Your network engineers will probably ask you to use a (mitm, therefore more secure) proxy.
We might even add this to core, when it is used in some places.

text/0000-manageable-http-proxies.md
timogoebel: I fully agree. Let's keep the smart proxies/capsule proxy settings in the proxy's settings.

johnpmitsch:
Discussed in the meeting, but I will ask here for other input, will there still be options to specify a http proxy in the installer or are we going to move away from that?

ehelms:
I'd be in favor of moving away from the installer.

PR #21, opened 2016-10-11
0018: Katello to Foreman Github Organization

ehelms:


text/0016-katello-as-a-plugin/0018-katello-to-foreman-github.md
```diff
@@ -0,0 +1,44 @@
+- Feature Name: Move Katello to Foreman Github organization
+- Start Date: 2016-10-11
+- RFC PR: (leave this empty)
+- Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Currently Foreman and Katello are split into two separate github organizations with all of Katello's gem's, puppet modules, installer and other code repositories.
```
daviddavis, line 9: s/gem's/gems/

text/0016-katello-as-a-plugin/0018-katello-to-foreman-github.md
```diff
@@ -0,0 +1,44 @@
...
+
+Nearly all Foreman code and plugins live centrally located at the Foreman github organization providing a single place to look for code repositories. The expectation is to move all Katello active code repositories to the Foreman github organization and deprecate using Katello's.
+
+# Detailed design
+[design]: #detailed-design
+
+The following are changes the Katello team feel would be required in order to merge into the Foreman github organization. These are in part due to the size of the Katello developer community as well as the size of the code set being moved over.
+
+ * someone(s) from the Katello organization should be owners on the Foreman organization
+ * move only *active* code repositories over
```
daviddavis, line 22: What happens to the inactive repositories? Also, the Katello github org?

text/0016-katello-as-a-plugin/0018-katello-to-foreman-github.md
```diff
@@ -0,0 +1,56 @@
...
+ * spec files in katello-packaging will require updates to point source tarballs from new locations
+ * puppet-katello_devel github links will need updating
+ * Puppetfile in katello-installer would need updating to point to new location
+ * Updates to prprocessor for repository locations
+ * prprocessor repository list updated
+ * Updates to CI jobs and scripts
+
+### Rename Katello
+
+An idea has been put forth to rename Katello's gem to match that of other plugin patterns. That is to move from `katello` to `foreman_katello` (and the CLI as well) by changing:
```
dLobatog, line 35: Not against it but why making the effort? I'm not sure I see any benefits

text/0016-katello-as-a-plugin/0018-katello-to-foreman-github.md
```diff
@@ -0,0 +1,56 @@
...
+ * RPM name
+
+But not changing:
+
+ * actual modules in code, i.e. no `Katello::` to `ForemanKatello::`
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This could cause some churn during the move of the repositories but in theory the Github redirects should lessen that. This presents a change to the current Foreman Github organization structure that may be controversial to some as adding owners is typically not a small thing.
```
dLobatog, line 48: It's not a big deal to add owners to Github these days thanks to teams. You can create a team per repo or per group of similar repos e.g - the ansible team contains maintainers for all ansible related repos.

text/0016-katello-as-a-plugin/0018-katello-to-foreman-github.md
ekohl, line 35: The benefit is that for all plugins that follow this pattern the puppet module can properly determine the correct package name. Not doing this means duplicating the OS-specific logic like in [puppet-katello](https://github.com/Katello/puppet-katello/blob/5de2e01795aac52e84cd88dc5d57b0841de8657e/manifests/params.pp#L4-L22) or the [puppet-foreman puppetdb plugin](https://github.com/theforeman/puppet-foreman/blob/6a42cf2f503791a1c1ea708bfcac3e5ea8fb1fb0/manifests/plugin/puppetdb/params.pp#L3-L33). It would be sufficient if the package name would be changed, but I think that the packaging guidelines want matching gem and package names. I have no opinion about the git repository name.

ehelms: There are plugins today in puppet-foreman that don't follow convention but still exist (e.g. foreman-tasks) ? How do they get around it?

ekohl: As you can see in the puppetdb example, it's a ton of duplication of logic.

text/0016-katello-as-a-plugin/0018-katello-to-foreman-github.md
ohadlevy: true, this is how all other plugins operate today, there are many sub teams currently on theforeman org.

PR #23, opened 2016-10-11
0020: Move Capsule to Smart Proxy

ehelms:


text/0016-katello-as-a-plugin/0020-remove-capsule.md
```diff
@@ -0,0 +1,48 @@
...
+
+# Summary
+[summary]: #summary
+
+A Capsule is essentially a Smart Proxy that deploys Pulp, and a reverse proxy to achieve content federation and an isolated communication point fo clients. Foreman deploys the Smart Proxy on port 8443 by default, while Katello deploys it on 9090 by default. Katello deployed Smart Proxies on 9090 because Candlepin was already running on 8443 on the server.
+
+# Motivation
+[motivation]: #motivation
+
+The "Capsule" terminology and code creates a perceived difference between a Foreman Proxy and the Capsule even though a Capsule is a Foreman Proxy with a few added features. Centralizing development of the Foreman Proxy will allow better innovation and less confusion for community users. The expected outcome is that we've removed the notion of a 'Capsule' and are centralized on the Foreman Proxy and adding features to it.
```
daviddavis, line 14: Are `Foreman Proxy` and `Smart Proxy` the same thing? Honestly asking here. If they are, I'd probably stick with one or use `Foreman Smart Proxy`, `Smart Proxy in Foreman`, etc

ehelms:
They are the same but I am honestly not clear what to call it either.

ares:
> They are the same but I am honestly not clear what to call it either.

I think smart proxy is used more frequently, also it's the [name](https://github.com/theforeman/smart-proxy) of the project on github.

ares:
I'm not thoroughly familiar with capsule but :+1: for what's written here.

GregSutcliffe:
They are indeed the same thing, and hsitorically both names have been used interchangably. From a purely personal view, I'd standardise on Foreman Proxy (including renaming the repo) because I find Smart Proxy to be a misleading name :p - but that's not for this RFC... :)

iNecas:
:+1: from technical point of view

text/0016-katello-as-a-plugin/0020-remove-capsule.md
```diff
@@ -0,0 +1,48 @@
...
+ * Add Smart Proxy features and plugins for:
+  - reverse proxy
+  - qpid-dispatch router
+ * Drop Pulp Node nomenclature for:
+  - Content
+  - Content Master
+ * Add functionality in puppet-capsule to puppet-foreman_proxy
+ * Rename capsule-remove to foreman-proxy-remove (or generic foreman-remove)
+ * Rename capsule-certs-generate to foreman-proxy-certs-generate
+ * Update server APIs to drop Capsule named APIs in favor of Smart Proxy APIs
```
jlsherrill, line 40: I assume we'd maintain backwards compatibility?

jlsherrill: Also, I'd add an item for 'updating hammer to use new apis'

dLobatog:
:+1: from me

text/0016-katello-as-a-plugin/0020-remove-capsule.md
lzap: Yes, this is confusing as well. Everything in the Foreman core is "Smart Proxy", but in packaging the file is named "Foreman Proxy". There is no easy way out, I'd suggest to create new official name "Foreman Smart Proxy" and use this in documentation and Web UI, so we can keep both, but it will "teach" users that this is essentially the same thing.

lzap:
Thank you for this effort! This will be huge improvement.

I also raised on the list to rename to "Foreman Smart Proxy" in the WebUI and Documentation, let's see how folks like it:

https://groups.google.com/forum/#!topic/foreman-dev/H-jbnDY5mhI

PR #25, opened 2016-10-11
0020: Converge Katello and Foreman Ports

ehelms:


text/0016-katello-as-a-plugin/0022-converge-katello-foreman-ports.md
```diff
@@ -0,0 +1,61 @@
...
+### Reverse Proxy
+
+The Katello reverse proxy deployed on Capsules to achieve isolation feature currently runs on 8443. This means that subscription-manager is configured to point at and go through 8443 and is yet another port that must be managed.
+
+ * Move reverse proxy to 443
+ * Maintain reverse proxy on 8443 and 443 for a release and then remove 8443
+
+### Custom Port Ranges
+
+One idea is to move all services to a set of well known ports and choose a port range that allows adding new services to a well defined location. For example:
```
sean797, line 39: Although this is a nice idea, I think this well create unnecessary pain for users upgrading -1

text/0016-katello-as-a-plugin/0022-converge-katello-foreman-ports.md
```diff
@@ -0,0 +1,61 @@
...
+
+ * Move to 8443: allows use of existing Smart Proxies and existing SELinux contexts
+ * Move to 9090: existing Capsules don't need to change, SELinux contexts must expose this port
+ * Move to 9070 (infrastructure port): SELinux contexts must be updated, centralized port range for all services to run on
+
+### Reverse Proxy
+
+The Katello reverse proxy deployed on Capsules to achieve isolation feature currently runs on 8443. This means that subscription-manager is configured to point at and go through 8443 and is yet another port that must be managed.
+
+ * Move reverse proxy to 443
```
ohadlevy, line 34: easy to break if there is any default apache ssl setup? we should not use common known ports.

text/0016-katello-as-a-plugin/0022-converge-katello-foreman-ports.md
```diff
@@ -0,0 +1,61 @@
...
+[motivation]: #motivation
+
+The difference in port usage causes differences in installations making it harder to add Katello to a Foreman, have common documentation and SELinux rules. The expectation is to converge on a common port usage paradigm that can be used for current services and as a guideline when adding new services in the future.
+
+# Detailed design
+[design]: #detailed-design
+
+### Capsule vs Smart Proxy Ports
+
+The Foreman Proxy uses 8443 by default on installation while the Capsule install for Katello uses 9090. This was originally done due to Candlepin's use of 8443; however, given Candlepin is a local installation the port it uses matters less.
```
ohadlevy, line 24: so in all in one default install, subscription manager clients connect directly to 8443(candlepin) or rather it get proxied via katello itself?

text/0016-katello-as-a-plugin/0022-converge-katello-foreman-ports.md
lzap: For what users? We need to change at least one port anyway, so some upgrade path must be there anyway.

I like the idea of doing in for once but correctly. And also it looks like `egrep ^90 /etc/services` is empty.

PR #26, opened 2016-10-11
0023: Converge Foreman and Katello adminstration scripts

ehelms:


text/0016-katello-as-a-plugin/0022-converge-katello-foreman-scripts.md
```diff
@@ -0,0 +1,72 @@
...
+ * foreman-tail
+ * foreman-rake
+ * foreman-config
+ * hammer_cli_foreman_admin
+
+Based on the current design of the scripts and where they live, the following design has been put forth:
+
+ * centralize adminstration into hammer_cli_foreman_admin
+  - this would allow a single repository to manage scripts
+  - this would allow users to have discoverability (e.g hammer admin --help to see list of optional commands)
```
ares, line 37: I like the idea, but it should be somehow noted that such commands can be run only on satellite server and ideally have some detection so if it's ran somewhere else, user would get clear error message. (`rpm -q $package` might be enough).

text/0016-katello-as-a-plugin/0022-converge-katello-foreman-scripts.md
```diff
@@ -0,0 +1,72 @@
+- Feature Name: Converge Foreman and Katello Adminstrative Scripts
+- Start Date: 2016-10-11
+- RFC PR: (leave this empty)
+- Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Between Foreman and Katello there are a variety of scripts that currently exist or emerging for adminstering the server and foreman proxy. For example, scripts that handle backup and restore, generating debug output, or managing services. As these scripts evolve, they can benefit from true code review, cnd entralization to increase innovation.
```
dLobatog: typo at the end `and centralization`

text/0016-katello-as-a-plugin/0022-converge-katello-foreman-scripts.md
```diff
@@ -0,0 +1,72 @@
...
+ * katello-debug
+ * katello-backup
+ * katello-restore
+ * katello-service
+ * katello-remove
+ * capsule-remove
+ * foreman-tail
+ * foreman-rake
+ * foreman-config
+ * hammer_cli_foreman_admin
```
dLobatog, line 31: katello-selinux-relabel, katello-certs-check, katello-certs-sign, foreman-selinux-relabel

text/0016-katello-as-a-plugin/0022-converge-katello-foreman-scripts.md
```diff
@@ -0,0 +1,72 @@
...
+
+Subcommands:
+ debug                         Generate debug output for upload
+ services                      Manage status of services
+ backup                        Generate server backup
+ restore                       Use previously generated backup to restore server
+ remove                        Remove all RPMs and configuration from the server
+ tail                          Tail all relevant logs
+
+``
```
sean797: s/``/```/

text/0016-katello-as-a-plugin/0022-converge-katello-foreman-scripts.md
```diff
@@ -0,0 +1,72 @@
...
+ * katello-service
+ * katello-remove
+ * capsule-remove
+ * foreman-tail
+ * foreman-rake
+ * foreman-config
+ * hammer_cli_foreman_admin
+
+Based on the current design of the scripts and where they live, the following design has been put forth:
+
```
sean797, line 34: Probably goes without saying but we should also leave the current scripts locations working with a deprecated message for at least one release.

text/0016-katello-as-a-plugin/0022-converge-katello-foreman-scripts.md
lzap: These are all in katello-selinux, I'd keep that package.

ehelms:
I updated this RFC based on comments (sorry if I missed any, it has been a while) and opened an initial PR to get some design discussion going.

iNecas:
I was looking into very similar context while looking into health-checking, remediation scripts and upgrade helpers. My additional requirements are:

1. zero-configuration: the tool should be able to detect what's running on the system and provide corresponding functionality based on that.
1. ability to use the same version of the tool against different versions of foreman/katello/satellite,
the tool will have definitions of the operations and will be able to determine what to do with the specific version.

This takes us to lower the dependencies on other things that are delivered as part of foreman/satellite (including even hammer).

I'm currently putting a POC of the code structure together, will have something to show next week.

text/0016-katello-as-a-plugin/0022-converge-katello-foreman-scripts.md
```diff
@@ -0,0 +1,74 @@
...
+ tail                          Tail all relevant logs
+
+```
+ 
+ * move existing scripts to hammer_cli_foreman_admin
+ * add maintainers of functionality to maintainers of hammer_cli_foreman_admin
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
```
iNecas, line 65: hammer is primarily used for taking to the foreman via API. This includes issues, such as:

1. slow start time: since there is already quite a lot of functionality, it takes more and more time to load (on my setup, 2 seconds just to load the whole thing. For now, I don't see too much benefits of building the tool as plugin into hammer, vs. standalone tool around clamp, limiting and focusing on maintenance tasks.

text/0016-katello-as-a-plugin/0022-converge-katello-foreman-scripts.md
```diff
@@ -0,0 +1,74 @@
...
+
+# Alternatives
+[alternatives]: #alternatives
+
+ 1) Use something other than hammer_cli_foreman_admin for CLI-ing scripts together like regular opt-parse
```
iNecas, line 70: +1 for 1),

text/0016-katello-as-a-plugin/0022-converge-katello-foreman-scripts.md
```diff
@@ -0,0 +1,74 @@
...
+Usage:
+    hammer [OPTIONS] SUBCOMMAND [ARG] ...
+
+Parameters:
+ SUBCOMMAND                    subcommand
+ [ARG] ...                     subcommand arguments
+
+Subcommands:
+ debug                         Generate debug output for upload
+ services                      Manage status of services
```
iNecas, line 52: I would rename this to health-checks

text/0016-katello-as-a-plugin/0022-converge-katello-foreman-scripts.md
```diff
@@ -0,0 +1,74 @@
...
+ backup                        Generate server backup
+ restore                       Use previously generated backup to restore server
+ remove                        Remove all RPMs and configuration from the server
+ tail                          Tail all relevant logs
```
iNecas, line 56: another administrative tasks:

1. upgrades - I don't think the current way of doing so inside installer hooks is a good approach
2. remediation scripts - fixes/suggestions for things we know might get wrong

text/0016-katello-as-a-plugin/0022-converge-katello-foreman-scripts.md
iNecas: +1 for katello-certs checks

text/0016-katello-as-a-plugin/0022-converge-katello-foreman-scripts.md
ehelms: I see this as two different pieces:

 1) hammer the code/library/integration with API
 2) hammer as THE CLI tool for Foreman

As a user, it is nice to be able to have one tool and let it tell you about the functionality you can do from an interface perspective (ignoring whats under the hood).

I suppose two CLI tools would be OK and limit user's required sphere of knowledge. For example, OpenShift has `oc` and `oadm`. However, I do find it annoying that you have to still figure out what you can do in one tool and what operations you do in the other tool.

text/0016-katello-as-a-plugin/0022-converge-katello-foreman-scripts.md
ehelms: I feel like they are two distinct pieces, solely because 'services' does more than 'check'. The current `katello-services` tool provides start, stop and restart abilities thus managing services. We could split out the status aspect.

ehelms:
Whichever route we go, I'd love to avoid the "plugin" style for administration tooling and scripts and have a single repository and tool for all functionality.

iNecas:
This complies with the design decisions we do: what we need to address is that the tool needs to be sensitive enough on what's present on the system it's run against and offer just the things that are relevant to the version of the system (upstream vs. downstream, plugins presence, registered proxies etc.)

text/0016-katello-as-a-plugin/0022-converge-katello-foreman-scripts.md
lzap: Yeah, I thought more on using hammer as a framework, integrating "hammer debug" into the foreman's "hammer" command was not good idea, it should have been perhaps separate command like this one.

What I find very useful is subcommands capabilities of hammer, this will be collections of subcommands essentially. If clamp (or whatever we choose to work with) provides that, that is fine I think.

In any case, this should be definitely a separate binary `foreman-maintain` instead of `hammer` (to be honest I think foreman-cli should be named `foreman-cli` from the day one).

lzap:
I am very happy we are building such a tool, I will do my part and rewrite foreman admin logging subcomand into this form once you have the first release and deprecate the old one.

text/0016-katello-as-a-plugin/0022-converge-katello-foreman-scripts.md
sideangleside: I see no issue in having a separate maintenance tool. I see hammer as how the operator interacts with Foreman _operationally_. Maintenance and/or debug tasks IMHO should always be in a separate tool, as there _will_ be tasks that can't be run remotely. Additionally, as an admin, I run hammer on my workstation. 

I dont want to have to remember 'commands that I can only run locally' versus 'commands that I can run remotely'

PR #31, opened 2016-12-19
0028 - popcon-style data agreggation for Foreman

GregSutcliffe:
I'll be referring to this in the draft community survey questions, which will go out to the dev-list tomorrow :)

### DIFF
--- a/text/0028-popcon-style-data-collection.md
+++ b/text/0028-popcon-style-data-collection.md
@@ -1,0 +1,172 @@
+- Feature Name: Popcon-style data aggregation for Foreman
+- Start Date: 2016-12-19
+- RFC PR: 
+- Issue: 
+
+# Summary
+[summary]: #summary
+
+Add an set of information to Foreman (similar to the About page) which will be
+sent (opt-out available) to the Foreman infra for aggregation. This enables us
+to learn more about things like userbase size, how up-to-date the community
+stays, plugin / CR popularity, smart-proxy features, and so on.
+
+# Motivation
+[motivation]: #motivation
+
+We have a lot of data on the Foreman developer community (via information from
+GitHub, Redmine, IRC, and the mailing lists). However, these people have
+self-selected to participate in our community, and represent a smaller, and
+more vocal, subset of the whole Foreman userbase. We know remarkably little
+about the wider community (not even how big it is).
+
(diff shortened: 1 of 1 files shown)


Klaas-:
I would also include aggregated information about the foreman infrastructure servers and the servers that foreman manages.
What OS is foreman running on. What kind of operating systems is foreman deploying in which versions. I don't think you should log exact numbers - in my mind it should be like <10 <100 <1000 ... or something like that.

I would combine that feature with an update notification - that could be an incentive to keep it turned on

text/0028-popcon-style-data-collection.md
```diff
@@ -0,0 +1,172 @@
+- Feature Name: Popcon-style data aggregation for Foreman
+- Start Date: 2016-12-19
+- RFC PR: 
+- Issue: 
+
+# Summary
+[summary]: #summary
+
+Add an set of information to Foreman (similar to the About page) which will be
```
sean797, line 9: s/an/a/

sean797 (commented):
I think having the aggregated data openly available to anyone with a link to it from the About page is important to show users what were collecting and how we are using it.

ekohl:
@GregSutcliffe talked about some options offline so I'll try to recollect some options we talked about and maybe some things we didn't talk about but just came up:

* Less specific information about the proxies

It may be sufficient to show the set of smart proxy versions and a set of all the features enabled. The exact distribution of those features can identify a setup while not adding a lot of information.
```json
{
  "proxies": {
    "count": "10",
    "versions": ["1.13", "1.14"],
    "features": ["dns", "puppet"]
  }
}
```

* Having an explicit setup phase

Using the new notification framework that landed in 1.15 it is possible to make it very clear it's happening. This can be used to tell it's opt-out or make it opt-in with a decent number of people still enabling it.

* Guarantee a certain opt-out period

When it's opt out then users should have at least some time before any data is sent. I'd propose a week. This can be combined with the explicit setup phase and a notification.

* Partial opt-in/opt-out

During this setup you can also ask for a level. Maybe you don't want to share anything about your proxies, but do want to share your foreman version. Statistics collection needs to handle "opting out" anyway since old versions won't report any options that get added after the initial introduction.

* Preview of the data that would be sent

As a user I would be more comfortable if you could see a preview of what would be sent.

PR #32, opened 2017-01-05
Improve readability of hammer help

tstrachota:


tstrachota:
@rohoover this is weaknesses of hammer's help pages in my point of view. May I ask you to review the solutions and add what you find as pain points?

Rohoover:
@tstrachota Sure thing, however I'm a little lost. Do you have a link or screenshot you can reference for me? Thanks!

text/0029-hammer-help-formatting.md
```diff
@@ -0,0 +1,260 @@
...
+```
+before:
+--organization ORGANIZATION_NAME         Organization name to search by
+--organization-id ORGANIZATION_ID        organization ID
+--organization-label ORGANIZATION_LABEL  Organization label to search by
+
+after:
+ --organization[-id|-label] ORGANIZATION  Organization
+```
+
```
ares, line 132: how bad idea or difficult to implement would it be to have single argument that would accept all?

```
--organization ORGANIZATION      Organization identifier, could be organization id, label, name
```

hammer would need to try to find the resource based on all which could generate 3 requests I guess. Looking at CLI for projects like openshift or openstack (I checked ironic only), they either use name only or have one parameter that accepts both name and UUID or have aliases that accept any of those.

text/0029-hammer-help-formatting.md
```diff
@@ -0,0 +1,260 @@
...
+--content-view-version requires --content-view[-id]
+```
+
+or c) below the option list as a more condensed text:
+```
+--product and --content-view requires --organization[-id]
+--content-view-version requires --content-view[-id]
+```
+
+The dependency descriptions can be generated from current apidocs.
```
ares, line 172: I prefer b) since I'd like to see what is also required when I use this specific option. a) + b) altogether would be nice

text/0029-hammer-help-formatting.md
```diff
@@ -0,0 +1,260 @@
...
+    --order ORDER                   Sort field and order, eg. ‘name DESC’
+    --page PAGE                     Page number, starting at 1
+    --per-page PER_PAGE             Number of results per page to return
+    --search SEARCH                 Filter results
+
+  -h, --help                        Print help
+```
+
+More complicated commands can benefit from option grouping and create custom groups. See example of
+host create command [below](#complete-examples).
```
ares, line 200: I like it :+1:

text/0029-hammer-help-formatting.md
tstrachota: Interesting idea. It would require more changes than just help but it's definitely doable and could be easier to use.

Two drawbacks I can see:

- It creates more requests.
- If you create a record called `"12"` (not that it would be wise) and there's already one with `id = 12` you're lost. But that might be a corner case we can live with.

tstrachota:
@Rohoover I'm sorry, I should have provided more background.

This rfc talks about improving readibility and information provided by help output of our command line  interface - `hammer`. Each command in `hammer` has a `--help` (or shortly `-h`) option that makes it print a page with description of the command and options it takes. I feel there's plenty of space for improvement in this area.

Since the output is all text, I didn't create any screenshots and just copy-pasted the text. Some example output (equivalent to screenshots in this case) is on the following pages:

- [hammer erratum list -h](https://github.com/tstrachota/rfcs/blob/c10c6c900212d62b6d0441d0cbaf961f2fc787fb/text/0029_hammer_help_formatting/erratum_list.md)
- [hammer host create -h](https://github.com/tstrachota/rfcs/blob/c10c6c900212d62b6d0441d0cbaf961f2fc787fb/text/0029_hammer_help_formatting/host_create.md)
- [hammer host list -h](https://github.com/tstrachota/rfcs/blob/c10c6c900212d62b6d0441d0cbaf961f2fc787fb/text/0029_hammer_help_formatting/host_list.md)
- [hammer product create -h](https://github.com/tstrachota/rfcs/blob/c10c6c900212d62b6d0441d0cbaf961f2fc787fb/text/0029_hammer_help_formatting/product_create.md)

I'm happy to do a G+ session with a bref demo of the cli or prepare a virtual machine where you can play with it if you find it helpful.

text/0029-hammer-help-formatting.md
ares: I was thinking of the possible collision too, I think hammer could detect it and ask user for clarification (I know it's even more work and requests :-) Anyway I'm not married to the idea, so let's see what others think. ` --organization[-id|-label]` is definitely acceptable from my side

Rohoover:
Thanks @tstrachota 

This reminds me of the YAML dump design I recently did. All that was really done was work the typography to make it more readable. See attached.

yaml

PR #33, opened 2017-02-22
RFC - subscriptions redesign

thomasmckay:
Place holder at the moment; will start with screenshots of existing, etc. to work from.

### DIFF
--- a/0000-subscriptions.md
+++ b/0000-subscriptions.md
@@ -1,0 +1,37 @@
+- Feature Name: Subscriptions Flow Redesign
+- Start Date: 2017-02-22
+- RFC PR: (leave this empty)
+- Issue: (leave this empty)
+
+# Summary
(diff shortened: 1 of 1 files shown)


waldenraines:
Is there some way we can remove the sync management page entirely and incorporate it into the products page(s), maybe using [patternfly's expanding list view](http://www.patternfly.org/pattern-library/content-views/list-view/#/api#list-view-with-expanding-rows)?

ohadlevy:
@waldenraines is content sync is in the scope of subscription redesign? I'm very much in favor of building a proper sync status (across proxies)

thomasmckay:
@ohadlevy - This was more focused on the initial import of the content provided by subscriptions. However, it would be interesting to see if the lifecycle environment / content view syncs to capsules could be represented in some way.

